PR #1, opened 2026-05-19
[ONBOARDING]docs: improve README install section

### LINKED ISSUE #2: pip install -e . fails on Ubuntu 24.04+ with externally-managed-environment
## Description

Running `pip install -e .` on a fresh Ubuntu VM (Multipass) fails with `externally-managed-environment` error. Starting from Ubuntu 23.04, PEP 668 marks the system Python as externally managed, preventing direct `pip install` without a virtual environment.

## Environment

- **OS:** Ubuntu 24.04 (Multipass VM on macOS host)
- **Python:** 3.14 (system)
- **pip:** system pip

## Error output

```
ubuntu@dev:~/odyssey$ pip install -e .
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.14/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation
or OS distribution provider. You can override this, at the risk of breaking
your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

## Solution

The README Install section should include virtual environment setup steps. This is good practice across all platforms, not just Ubuntu.

```bash
python3 -m venv .venv
source .venv/bin/activate
pip install -e .
```

### LINKED ISSUE #3: pip install -e '.[all]' fails on Linux: evdev needs build-essential and python3-dev
## Description

Running `pip install -e ".[all]"` on a fresh Linux VM fails because `evdev` (a transitive dependency via `robosuite → pynput → evdev`) needs to compile a C extension and the system lacks `gcc` and Python dev headers.

**Dependency chain:** `robosuite >= 1.4` → `pynput` → `evdev` → requires C compilation

## Environment

- **OS:** Ubuntu 22.04 (GCP Deep Learning VM, `g2-standard-8` with NVIDIA L4)
- **Python:** 3.10.12 (system)
- **Image:** `common-cu129-ubuntu-2204-nvidia-580`

## Error output

```
building 'evdev._input' extension
creating build/temp.linux-x86_64-3.10
creating build/temp.linux-x86_64-3.10/evdev
x86_64-linux-gnu-gcc ... -c evdev/input.c -o build/temp.linux-x86_64-3.10/evdev/input.o -std=c99
error: command 'x86_64-linux-gnu-gcc' failed: No such file or directory
```

Additionally, `build-essential` is not found without running `sudo apt update` first on a fresh VM.

## Solution

Document system prerequisites for Linux in the README before the install commands:

```bash
# Linux only — needed to compile native extensions
sudo apt update && sudo apt install build-essential python3.10-dev -y
```

## Notes

- Does **not** affect the base install (`pip install -e .`) — only `.[all]` via the `robosuite` extra.
- Does **not** affect macOS (no `evdev` dependency on Darwin).
- Was not caught during Multipass testing because only the base install was tested there.

### LINKED ISSUE #4: OpenVLA quickstart: missing draccus dependency
## Summary

Running the OpenVLA quickstart fails because the upstream `openvla/vla-scripts/finetune.py` requires `draccus` for config parsing, but `draccus` is not listed in the `[project.optional-dependencies].openvla` extra in `pyproject.toml`.

The README correctly documents cloning the repo and setting `OPENVLA_REPO_PATH`, but even after `pip install -e ".[all,dev]"`, the `draccus` module is missing.

## Steps to reproduce

1. `pip install -e ".[all,dev]"`
2. `git clone https://github.com/openvla/openvla ~/openvla`
3. `export OPENVLA_REPO_PATH=~/openvla`
4. `python ~/openvla/vla-scripts/finetune.py --help`

```
ModuleNotFoundError: No module named 'draccus'
```

## Root cause

`pyproject.toml` declares:

```toml
openvla = [
    "transformers>=4.40",
    "peft>=0.8",
    "torch>=2.1",
]
```

`draccus` is missing from this list. It is a dependency of the upstream OpenVLA finetune script, not of Odyssey itself, but since Odyssey launches that script in the same virtualenv, it needs to be available.

## Proposed fix

Either:

- **Option A**: Add `draccus` to the `openvla` extra in `pyproject.toml`.
- **Option B**: Document that users should also run `pip install -e ~/openvla` to pull in all upstream OpenVLA dependencies (this would install draccus transitively).
- **Option C**: Both.

SoyGema:
## Summary

- Add `[all]` extra to `pyproject.toml` that bundles `huggingface`, `openvla`, and `robosuite` for a single-command install of the full training + evaluation stack.
- Restructure the README Install section into three clear tiers: base (lightweight CLI), `.[all]` (real runs), and `.[all,dev]` (contributor tooling).
- Update git clone URL and directory name to point at `lovellai-dev/odyssey`.
- Keep individual extras documented as a secondary note for users who only need part of the stack.

## Test plan

- [x] `pip install -e .` installs successfully
- [x] `pip install -e ".[all]"` resolves and installs all runner/provider deps
- [x] `pip install -e ".[dev]"` installs dev tooling
- [x] `odyssey --help` works after base install
- [x] Smoke test section (`odyssey validate`, `odyssey run --use-mock-runner`) — in progress

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,8 +150,8 @@ one agent — will become possible when the agent cap lifts.
 ## Install
 
 ```bash
-git clone https://github.com/femtechie/lovell-odyssey.git
-cd lovell-odyssey
+git clone https://github.com/lovellai-dev/odyssey.git
+cd odyssey
 pip install -e .
 ```
 
--- a/README.md
+++ b/README.md
@@ -162,17 +162,10 @@ run `validate`, `list`, `status`, and `run --use-mock-runner` against any
 mission spec without a GPU. `.[all]` adds everything needed for real training
 and evaluation runs.
 
-You can also install extras individually if you only need part of the stack:
-
-```bash
-pip install -e ".[huggingface]"   # HF model + dataset providers
-pip install -e ".[openvla]"       # OpenVLA training runner deps
-pip install -e ".[robosuite]"     # Robosuite evaluation runner deps
-```
-
 ## 60-second smoke test (no GPU, no network)
 
 ```bash
+# Validate the mission spec
 $ odyssey validate examples/quickstart-openvla/mission.yaml
 OK  examples/quickstart-openvla/mission.yaml
   spec version : 0.1
@@ -180,6 +173,7 @@ OK  examples/quickstart-openvla/mission.yaml
   robot        : embodiment=franka_panda
   tasks        : 1 training, 1 evaluation
 
+# Run the full mission with a CPU mock (no GPU needed)
 $ odyssey run examples/quickstart-openvla/mission.yaml --use-mock-runner
 {"ts": "...", "event": "mission.created", ...}
 {"ts": "...", "event": "mission.queued", ...}
@@ -190,9 +184,11 @@ $ odyssey run examples/quickstart-openvla/mission.yaml --use-mock-runner
 COMPLETED  c1756bad855e45cc9a95b5b0566c948b
   overall_grade : 1.000
 
+# List all missions from the local DB
 $ odyssey list
 c1756bad855e  COMPLETED   openvla-bridge-lift   2026-05-17T23:18:34+00:00  grade=1.000
 
+# Show details for a specific mission (prefix match)
 $ odyssey status c1756bad
 COMPLETED  c1756bad855e45cc9a95b5b0566c948b
   name         : openvla-bridge-lift
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ artifacts under `~/.odyssey/runs/<mission-id>/<task-id>/`.
 
 ## What it is
 
-You describe a mission in YAML — a robot, a model, a dataset to train on, an
+You describe a mission in YAML — a robot, an agent, a dataset to train on, an
 evaluation benchmark to score against — and `odyssey run` walks it through the
 full lifecycle: load → validate → execute training tasks → execute the
 evaluation task → persist results. Local-mode by default; the hosted Lovell
--- a/README.md
+++ b/README.md
@@ -149,6 +149,12 @@ one agent — will become possible when the agent cap lifts.
 
 ## Install
 
+**Linux only** — install build dependencies before proceeding (needed by `.[all]`):
+
+```bash
+sudo apt update && sudo apt install build-essential python3-dev -y
+```
+
 ```bash
 git clone https://github.com/lovellai-dev/odyssey.git
 cd odyssey
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ artifacts under `~/.odyssey/runs/<mission-id>/<task-id>/`.
 
 ## What it is
 
-You describe a mission in YAML — a robot, an agent, a dataset to train on, an
+You describe a mission in YAML — a robot, a model, a dataset to train on, an
 evaluation benchmark to score against — and `odyssey run` walks it through the
 full lifecycle: load → validate → execute training tasks → execute the
 evaluation task → persist results. Local-mode by default; the hosted Lovell
--- a/README.md
+++ b/README.md
@@ -152,6 +152,8 @@ one agent — will become possible when the agent cap lifts.
 ```bash
 git clone https://github.com/lovellai-dev/odyssey.git
 cd odyssey
+python3 -m venv .venv
+source .venv/bin/activate
 pip install -e .              # CLI, validate, mock runs (lightweight)
 pip install -e ".[all]"       # real training + evaluation (torch, robosuite…)
 pip install -e ".[all,dev]"   # + pytest, ruff, mypy
--- a/README.md
+++ b/README.md
@@ -1,16 +1,13 @@
 <h1 align="center">Lovell Odyssey</h1>
 
-[![CI](https://github.com/lovellai-dev/odyssey/actions/workflows/ci.yml/badge.svg?branch=main)](https://github.com/lovellai-dev/odyssey/actions/workflows/ci.yml)
-[![Python 3.10+](https://img.shields.io/badge/python-3.10%2B-blue.svg)](https://www.python.org/downloads/)
-[![License: Apache 2.0](https://img.shields.io/badge/license-Apache%202.0-green.svg)](LICENSE)
-[![Status: Pre-Alpha](https://img.shields.io/badge/status-pre--alpha-orange.svg)]()
+<p align="center">
+  <a href="https://github.com/lovellai-dev/odyssey/actions/workflows/ci.yml"><img src="https://github.com/lovellai-dev/odyssey/actions/workflows/ci.yml/badge.svg?branch=main" alt="CI"></a>
+  <a href="https://www.python.org/downloads/"><img src="https://img.shields.io/badge/python-3.10%2B-blue.svg" alt="Python 3.10+"></a>
+  <a href="LICENSE"><img src="https://img.shields.io/badge/license-Apache%202.0-green.svg" alt="License: Apache 2.0"></a>
+  <img src="https://img.shields.io/badge/status-pre--alpha-orange.svg" alt="Status: Pre-Alpha">
+</p>
 
 Open-source framework for defining, running, and benchmarking robot training missions.
-
-> [!WARNING]
-> **Pre-alpha (v0.0.x).** Not yet on PyPI. The first public alpha is
-> targeted at `v0.1.0-alpha`. API, CLI, schemas, and wire protocols are
-> subject to change without notice.
 
 ## Install
 
--- a/README.md
+++ b/README.md
@@ -7,17 +7,18 @@
 
 Open-source framework for defining, running, and benchmarking robot training missions.
 
-> **Status: pre-alpha (v0.0.x).** Not yet on PyPI. The first public alpha is
-> targeted at `v0.1.0-alpha`. The API, CLI, schemas, and wire protocols are
-> still subject to change without notice.
+> [!WARNING]
+> **Pre-alpha (v0.0.x).** Not yet on PyPI. The first public alpha is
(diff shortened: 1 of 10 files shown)


SoyGema:
## README restructure + shields.io badges

This push includes two changes:

**1. Shields.io badges** — CI status, Python version, license, and pre-alpha status badges added to the top of the README.

**2. README restructured for faster onboarding.** The previous layout buried Install at line ~150 behind ~130 lines of architecture concepts (Missions, Robots, Agents, Robot specs). A newcomer landing on the repo had to scroll through dense theory before they could even clone and run.

The new order follows the pattern used by high-traffic OSS repos (e.g. `huggingface/transformers`):

> title → badges → one-liner → install → quick start → what it is → CLI reference → project layout → advanced usage → status → license

Install is now reachable within the first 15 lines.

**No content was deleted** — the full Concepts section (missions, robots and agents, robot specs in v0.0.x) was moved verbatim to `docs/concepts.md` and linked from the README.

docs/concepts.md
```diff
@@ -0,0 +1,131 @@
+# Concepts
+
```
femtechie, line 2: The reorganization makes sense. Nice!

docs/concepts.md
```diff
@@ -0,0 +1,131 @@
...
+it runs the robot — that is, it composes the current checkpoints of
+every agent in the loadout.
+
+**Today, Odyssey fine-tunes the underlying model for one of the
+robot's agents at a time.** v0.0.x enforces exactly one agent on the
+robot (the implicit PILOT), and the eval composes a single policy from
+that single agent. Multi-agent loadouts (PILOT + one or more
+SPECIALISTs) and the multi-agent execution that goes with them ship
+when the multi-agent runtime lands. What v0.0.x does *not* model from
+the brain paper — agent persona / goals / success criteria,
```
femtechie: Suggest not to mention the "brain paper" here since it will not likely be fully made public.

docs/overview.md
```diff
@@ -0,0 +1,40 @@
...
+- Optional extras: `torch` + `transformers` + `peft` (OpenVLA training), `robosuite` + `mujoco` (evaluation), `huggingface-hub` + `datasets`
+- Tooling: hatchling build, ruff linter, mypy strict, pytest
+
+## CLI
+
+`odyssey init`, `validate`, `run`, `list`, `status` — with `--use-mock-runner` for GPU-free testing on a laptop.
+
+## Focus
+
+The project is oriented toward **VLA (Vision-Language-Action) model** fine-tuning and robotic sim evaluation, with OpenVLA as the first supported model and Robosuite as the first eval environment.
```
femtechie: The pre-alpha version supports VLA only. We will merge multi-agent training in a future release, which will also support VLMs.

README.md
```diff
@@ -202,26 +72,17 @@ COMPLETED  c1756bad855e45cc9a95b5b0566c948b
...
-```bash
-pip install -e ".[huggingface,openvla,robosuite]"
-git clone https://github.com/openvla/openvla.git /srv/openvla
-
-odyssey run examples/quickstart-openvla/mission.yaml
-```
+## What it is
 
-Hardware: 24 GB GPU (RTX 4090-class or better) for the OpenVLA fine-tune.
+You describe a mission in YAML — a robot, a model, a dataset to train on, an
```
femtechie: You describe a mission in YAML - a robot, **an agent**, and a dataset to train on..."

femtechie (changes requested):
This is a good improvement on the docs. Added a few comments, include a couple of corrections.

SoyGema (commented):
changed in PR , can´t change them in this branch due to protection

docs/concepts.md
SoyGema: changed in https://github.com/lovellai-dev/odyssey/pull/12

README.md
SoyGema: excellent catch ! dressed in https://github.com/lovellai-dev/odyssey/pull/12

docs/concepts.md
```diff
@@ -0,0 +1,125 @@
...
+In the Lovell AI architecture, a **robot** is more than an embodiment —
+it's a composition of an embodiment with a **loadout of agents**. Every
+robot has exactly one **PILOT** (running a Vision-Language-Action model
+with physical authority over the actuators) and zero or more
+**SPECIALISTs** (running language models for delegated reasoning — map
+queries, calculations, lookups). Each agent is authored with a persona,
+goals, and success criteria, and runs against a pinned model checkpoint;
+its behavior is conditioned at runtime by materialized artifacts
+produced from that prose. The fuller picture is described in the Lovell
+AI robot-brain paper.
```
femtechie: Scrub robot brain paper as the document is not published.

docs/overview.md
```diff
@@ -0,0 +1,40 @@
+# Lovell Odyssey Overview
+
+An **open-source framework for defining, running, and benchmarking robot training missions**. Built by Lovell AI, currently in pre-alpha (v0.0.1).
+
+You describe a mission in YAML — specifying a robot, a model, a dataset, and an evaluation benchmark — and `odyssey run` orchestrates the full lifecycle: load → validate → train → evaluate → persist results.
```
femtechie: Potential regression - "specifying a robot, a model, a dataset, and an evaluation benchmark..." seems to have reverted to the change that read "specifying a robot, an agent, a dataset, and an evaluation benchmark.

docs/overview.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Key Concepts
+
+- **Missions** — YAML specs that define what to train, how to evaluate, and on which robot
```
femtechie, line 9: Consider if "Agents" should be added as a "key concept".

femtechie (changes requested):
Good clarifications. I was able to install and run the framework without GPU.

docs/concepts.md
SoyGema: Re: https://github.com/lovellai-dev/odyssey/pull/1#discussion_r3313860292 — Scrubbed all "robot-brain paper" references from the entire codebase (docs and source docstrings). See commit https://github.com/lovellai-dev/odyssey/commit/15feb2fd2b9c60a1d847e477b0a74ab9d52b3a08.

docs/overview.md
SoyGema: After careful thought  , **model** might be precise for this context.  The white paper stands for an hierarchy that is clear and concise.  Agents refers to the PILOT and/or STRATEGY entities.  I would keep model instead of agent, and in next iteration, I would conceptually introduce the brain vision . What do you think about this?

docs/overview.md
SoyGema: TBH all the brain vision should be carefully be here . I think however that we can leave it for next iteration. Compositional brain arch and agents might be wanted to be described jointly . 

If you want to include the concept Agent now, what I would suggest is something as

`Agent — The autonomous entity resulting from training a mission. It becomes part of the robot as a PILOT or SPECIALIST.`

What do you think about this?

README.md
SoyGema: I´m going to rollback to **model** here due to our discussion . Now 9c35d58

Would you like to introduce the concept Agent here ? Like

`  You train an agent by defining a mission in YAML`

LMK your thoughts

SoyGema: Done! The line now reads:

> You train an agent by describing a mission in YAML — a robot, a model, a dataset to train on…

This way we introduce the concept of **agent** (the outcome of training) while keeping **model** (the neural network being fine-tuned) — which I think captures both perspectives from our discussion nicely.

We also added both **Models** and **Agents** as key concepts in `docs/overview.md`, using your suggested definition for Agent:

> **Agents** — The autonomous entity resulting from training a mission. It becomes part of the robot as a PILOT or SPECIALIST

Would you mind taking another look when you get a chance? Really appreciate your thorough reviews — they've made the docs much stronger. Thank you!

docs/overview.md
SoyGema: Addressed — the line now reads "You train an agent by describing a mission in YAML — specifying a robot, a model, a dataset…". This keeps **model** for the technical context (what's being fine-tuned) while introducing **agent** as the higher-level outcome, as we discussed. Thank you for flagging the regression\!

docs/overview.md
SoyGema: Great suggestion — done! Added both **Models** and **Agents** as key concepts, using the definition you proposed:

- **Models** — The neural network being trained (e.g. OpenVLA). A mission specifies which model to fine-tune and how
- **Agents** — The autonomous entity resulting from training a mission. It becomes part of the robot as a PILOT or SPECIALIST

Fully agree that the compositional brain architecture deserves a proper treatment in a future iteration. For now this gives readers the essential vocabulary without overloading the overview. Thank you for the thoughtful suggestion!

SoyGema (commented):
Hi! All three open threads have been addressed in the latest commits. Summary of changes:

- **"model" vs "agent"**: now reads "You train an agent by describing a mission in YAML" — introduces agent as the outcome while keeping model for the technical context
- **Agents as key concept**: added both **Models** and **Agents** to Key Concepts in overview.md, using your suggested definition
- **Bonus**: renamed "Lovell Odyssey" → "Odyssey" across README, overview, and code docstrings; centered the tagline

Would really appreciate another look when you have a moment — thank you so much for all the detailed feedback, it has genuinely improved the docs!

femtechie (approved):
Thanks for these updates. Looks complete.

SoyGema:
Merging stuff! 
daftpunk-tron

PR #6, opened 2026-05-21
[TRAINING]fix(runners): torchrun support for OpenVLA + GCP compatibility

SoyGema:
## Summary

- Add `use_torchrun` / `torchrun_nproc` fields to `TrainingProcessSpec` so runners can launch via `torchrun` instead of `python`
- Enable `use_torchrun=True` for the OpenVLA runner (required for DDP initialization)
- Filter Odyssey-only config keys (`method`, `lora_alpha`, `epochs`) from CLI args passed to upstream `finetune.py`
- Add `dataset_name: bridge_orig` to quickstart mission (OpenVLA requires a valid OXE dataset name)

## Context

Discovered during end-to-end testing on a GCP VM (NVIDIA L4, 24GB). Full details in #5.

### Key findings:
- OpenVLA's `finetune.py` requires `torchrun` (not `python`) for DDP initialization
- On GCP single-GPU VMs, `export NCCL_NET=Socket` is required to bypass Google's custom NCCL plugin (`gIB`)
- OpenVLA expects RLDS/TFDS format datasets, not LeRobot/HuggingFace format
- The `--dataset_name` must be a valid OXE name (e.g. `bridge_orig`), not the Odyssey task name

### Verified pipeline (on GCP L4):
- ✅ Model download from HuggingFace (`openvla/openvla-7b`, ~15GB)
- ✅ Model loading + LoRA setup (27M/7.5B trainable params)
- ✅ DDP initialization (with `NCCL_NET=Socket`)
- ✅ RLDS dataset discovery and reading
- ⏳ End-to-end training (blocked on full dataset download, in progress)

## Scope & limitations

This PR targets **single-node execution only**. The launcher uses `torchrun --standalone`, which assumes all GPUs are on one machine.

| Scenario | Status |
|----------|--------|
| Single GPU | ✅ Works (current default: `nproc_per_node=1`) |
| Multi-GPU, single node | ⚠️ Requires exposing `torchrun_nproc` to mission config (field exists on `TrainingProcessSpec` but is not yet configurable from YAML — currently hardcoded to 1) |
| Multi-node (multiple machines) | ❌ Not supported. Would require replacing `--standalone` with `--nnodes`, `--node_rank`, `--master_addr`, `--master_port`, and an orchestration layer to launch processes across machines |

### Future work for multi-node support

To support multi-node training, `TrainingProcessSpec` and `run_training_subprocess` would need to:
[...]

Issue #5, opened 2026-05-20
OpenVLA runner: torchrun launch + invalid CLI args passed to finetune.py


SoyGema:
## Summary

The OpenVLA runner has two bugs that cause `finetune-openvla` to fail at runtime:

### Bug 1: subprocess uses `python` instead of `torchrun`

The upstream `openvla/vla-scripts/finetune.py` wraps the model with `torch.nn.parallel.DistributedDataParallel` (DDP) unconditionally (line 185). DDP requires `torch.distributed` to be initialized, which only happens when launched via `torchrun` (or equivalent).

**Current** (`src/odyssey/runners/subprocess.py` line 86):
```python
cmd = ["python", spec.script_path, *spec.argv_extra]
```

**Error**:
```
ValueError: Default process group has not been initialized,
please make sure to call init_process_group.
```

**Fix**: The subprocess runner needs a `use_torchrun` option. For single-GPU, the launch command should be:
```
torchrun --standalone --nproc_per_node=1 vla-scripts/finetune.py <args>
```

### Bug 2: mission.yaml config keys passed as invalid CLI args

`build_openvla_argv()` in `src/odyssey/runners/openvla.py` iterates over all keys in `task.config` and passes them through as `--key value` to the finetune script. The quickstart mission.yaml (`examples/quickstart-openvla/mission.yaml`) defines:

```yaml
config:
  method: peft_lora       # NOT accepted by finetune.py
  lora_rank: 8            # valid
  lora_alpha: 16          # NOT accepted by finetune.py
  batch_size: 2           # valid
  learning_rate: 5.0e-5   # valid
  epochs: 1               # NOT accepted by finetune.py (uses --max_steps)
```

The `handled` set (line 165) only filters `vla_path`, `data_root_dir`, `dataset_name`. So `method`, `lora_alpha`, and `epochs` get passed as `--method peft_lora --lora_alpha 16 --epochs 1`, which draccus rejects with exit code 2.

**Valid args** per `finetune.py --help`:
[...]

PR #7, opened 2026-05-21
[TRAINING]fix(mission): add dataset_name to OpenVLA quickstart mission

SoyGema:
> [!IMPORTANT]  
> This PR is NOT to develop nor to main , it is to another fix branch
> RULES were activated while I was working on that branch .

Merging this branch will allow me to work in the fix.
Thanks 

## Summary

Adds `dataset_name: bridge_orig` to the OpenVLA quickstart mission config.

## Problem

The OpenVLA runner (`openvla.py:153`) builds the `--dataset_name` CLI arg like this:

```python
argv += ["--dataset_name", str(config.get("dataset_name", task.name))]
```

When `dataset_name` is not in the mission config, it falls back to `task.name` which is `finetune-openvla`. This is **not a valid OXE dataset name** and causes:

```
KeyError: 'finetune-openvla'
```

OpenVLA's `finetune.py` requires the dataset name to match a key in the Open X-Embodiment (OXE) config registry (`prismatic/vla/datasets/rlds/oxe/configs.py`).

## Fix

Add `dataset_name: bridge_orig` explicitly to the mission.yaml config. `bridge_orig` is the OXE name for Bridge V2 in RLDS/TFDS format.

## Important: dataset format

OpenVLA expects the dataset in **RLDS/TFDS format**, not LeRobot/HuggingFace format. The mission.yaml still references `lerobot/bridge_v2` as the dataset source, but the actual data must be downloaded separately in RLDS format:

```bash
wget -r -nH --cut-dirs=4 --reject="index.html*" \
  https://rail.eecs.berkeley.edu/datasets/bridge_release/data/tfds/bridge_dataset/
mv bridge_dataset bridge_orig
```

Then `--data_root_dir` must point to the parent directory containing `bridge_orig/`. The full dataset is ~124 GB.

This format mismatch is a known limitation tracked in #5.

## Context

Discovered during end-to-end testing on GCP VM (NVIDIA L4). Part of the work in PR #6.

### DIFF
--- a/examples/quickstart-openvla/mission.yaml
+++ b/examples/quickstart-openvla/mission.yaml
@@ -47,6 +47,7 @@ tasks:
       batch_size: 2
       learning_rate: 5.0e-5
       epochs: 1
+      dataset_name: bridge_orig  # OXE name required by OpenVLA's finetune.py
 
   - name: eval-on-robosuite-lift
     kind: evaluation


SoyGema:
Superseded by #10, which includes this commit plus the buffer fix and real-time subprocess output streaming.

examples/quickstart-openvla/mission.yaml
```diff
@@ -47,6 +47,7 @@ tasks:
       batch_size: 2
       learning_rate: 5.0e-5
       epochs: 1
+      dataset_name: bridge_orig  # OXE name required by OpenVLA's finetune.py
```
femtechie, line 51: Should the OXE key be passed into the runner at runtime?

SoyGema: So this is what I´ve learnt so far :
OXE = Open X-Embodiment, standard/registry of multiembodiment robotics dataset. **OKE Key** (in this case bridge_orig) is a register identifier that finetune.py uses for : 
1.  Search dataset configuration inside internal OXE register.
2. Build real dataset rute :  {data_root_dir}/{dataset_name}/1.0.0/
3. Select data loader and correct format handler.

Why this didn´t influence runtime?

 If we look at openvla.py:153:
  argv += ["--dataset_name", str(config.get("dataset_name", task.name))]

` dataset_name ` always is paased to the subprocess as CLI argument `--dataset_name`. 
It is not that it doesn´t influence it, it is just that this value only has effect inside finetune.py script from OpenVLA subprocess , NOT in Odyssey logic. 



So yeah, I understand that passing this argument is not necessary, let me please think of a more inteligent workaround that might work

PR #10, opened 2026-05-25
[TRAINING] feat(runners): stream subprocess output + buffer fix + dataset_name

SoyGema:
## Summary

- Stream subprocess stdout/stderr directly to the terminal in real-time (tqdm bars, model loading, training output all visible now)
- Switch from `readline()` to `read(8192)` chunks to handle tqdm `\r` output without blocking
- Increase asyncio buffer to 10 MB as safety net
- Add `dataset_name: bridge_orig` to OpenVLA quickstart mission config

Replaces PR #9 (`fix/subprocess-buffer`).

## Problem

Users running `odyssey run` saw `dataset_loading` and then silence — no training output visible. Root causes:
1. Subprocess stdout went only to `logger.info()`, not shown in CLI
2. `readline()` blocks on tqdm `\r` output (no `\n` terminator)

## Fix

Write raw bytes from subprocess directly to `sys.stdout.buffer` for real-time output, while still parsing `\n`-terminated lines for structured progress events.

## Context

Part of #5. Discovered during end-to-end testing on GCP VM (NVIDIA L4).

## Test plan

- [ ] `odyssey run examples/quickstart-openvla/mission.yaml` shows live training output
- [ ] tqdm progress bars render correctly
- [ ] Structured events (`training_step`, `checkpoint_save`) still emitted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #11, opened 2026-05-25
fix(mission): add data_root_dir to OpenVLA quickstart

SoyGema:
## Summary

Add `data_root_dir` to the OpenVLA quickstart mission config.

## Problem

Without `data_root_dir`, the OpenVLA runner falls back to `task.dataset.ref` (`lerobot/bridge_v2`) — an HF reference, not a local path. TFDS then fails with:

```
DatasetNotFoundError: Dataset bridge_orig not found.
```

## Fix

Add `data_root_dir: /path/to/dataset` as a placeholder. Users must set this to the parent directory containing `bridge_orig/1.0.0/`.

## Context

Discovered during end-to-end `odyssey run` testing on GCP VM. Part of #5.

### DIFF
--- a/examples/quickstart-openvla/mission.yaml
+++ b/examples/quickstart-openvla/mission.yaml
@@ -47,7 +47,8 @@ tasks:
       batch_size: 2
       learning_rate: 5.0e-5
       epochs: 1
-      dataset_name: bridge_orig  # OXE name required by OpenVLA's finetune.py
+      dataset_name: bridge_orig      # OXE name required by OpenVLA's finetune.py
+      data_root_dir: /path/to/dataset  # parent dir containing bridge_orig/1.0.0/
 
   - name: eval-on-robosuite-lift
     kind: evaluation


SoyGema:
https://github.com/lovellai-dev/odyssey/pull/7#discussion_r3301733654

SoyGema (commented):
The real problem:  Missions dataset LERobot is not accepted by OpenVLA , and overriding with config is not clean nor a good design decision. 

 current mission says:

```
  dataset:                              # ← "This is my dataset"
    source: huggingface
    ref: lerobot/bridge_v2              #    HuggingFace, LeRobot format
    partial: 1000
    format: lerobot
  config:
    dataset_name: bridge_orig           # ← "Actually, use THIS dataset"
    data_root_dir: /path/to/dataset     #    RLDS/TFDS format, on local disk
```

  The dataset block declares a HuggingFace LeRobot dataset (`lerobot/bridge_v2`). But OpenVLA cannot use LeRobot format — it requires RLDS/TFDS format. So the `config block` overrides everything with a completely different dataset location
  and format. The dataset block is effectively dead code — the runner ignores it in favor of config.data_root_dir and config.dataset_name.

  This is the contradiction and the lack of clarity @femtechie spotted.. The `dataset_name `(OXE key) is not a hyperparameter — it's a dataset identity. It shouldn't live in config next to `batch_size `and `learning_rate`.


CLEAN FIX:;

Use another name for  ecosystem : 
it's the same dataset, different format.  The key fact: OpenVLA can only consume RLDS format. Its entire data pipeline (prismatic → tfds.builder()) is built around TFDS. It cannot read LeRobot format at all.


  ```
dataset:
    source: oxe
    ref: bridge_orig        # ← OXE key IS the ref (identity)
    format: rlds
  config:
    data_root_dir: /path/to/dataset   # deployment path (or future: env var / provider)
    batch_size: 2
    learning_rate: 5.0e-5
    epochs: 1
```

SoyGema:
Superseded: moving dataset identity (OXE key + data_root_dir) out of config and into the dataset spec block. New PR will branch from #10.

PR #15, opened 2026-06-10
feat: OpenVLA training pipeline fixes + evaluation wiring + telemetry alignment

### LINKED ISSUE #5: OpenVLA runner: torchrun launch + invalid CLI args passed to finetune.py

SoyGema:
## Summary

End-to-end fixes for the OpenVLA training pipeline and initial wiring of the evaluation path, plus telemetry improvements that align Odyssey's event contract with lai-trainer.

### Training pipeline (working on GPU VM)
- Launch OpenVLA fine-tuning via `torchrun` instead of plain `python` (multi-GPU ready)
- Filter Odyssey-only config keys (`method`, `lora_alpha`, `epochs`) so they don't leak to upstream `finetune.py`
- Switch dataset from HuggingFace/lerobot to OXE/RLDS — move the OXE registry key (`bridge_orig`) to `dataset.ref` as single source of truth
- Add `data_root_dir` and eval `unnorm_key` to quickstart mission spec

### Subprocess output / verbosity
- Increase asyncio subprocess buffer from 64 KB to 10 MB to prevent `LimitOverrunError` with tqdm's `\r`-based progress bars
- Stream raw subprocess bytes to terminal in real-time via `read(8192)` + `sys.stdout.buffer.write()` so the developer sees tqdm live
- Internal line buffer accumulates partial chunks for structured event parsing

### Evaluation pipeline (code-complete, not yet tested on VM)
- Auto-wire `make_openvla_policy()` in `RobosuiteRunner` when no custom `policy_factory` is injected
- Add `_make_eval_env()` with camera enabled (`use_camera_obs=True`) for image observations
- Support both LoRA adapter checkpoints and full merged model checkpoints
- Resolve `checkpoint-NNN/` subdirectories inside OpenVLA run dirs
- Remove duplicate `openvla_policy.py` — all policy code consolidated in `openvla.py`

### Telemetry alignment with lai-trainer
- Add `ProgressEvent` Pydantic model formalizing the progress event contract (same fields as lai-trainer's schema)
- Add monotonic `seq` counter per task on `TaskContext.emit_progress()` for ordered replay and future polling
- Validate progress payloads at emission time instead of passing ad-hoc dicts

### Housekeeping
- Add `docs/migration/` to `.gitignore` (internal analysis docs)

## Test plan
- [x] 181 unit tests pass locally
- [x] Training pipeline tested with multiple runs on GCP L4 GPU (500 steps at ~1.49 it/s, `save_steps: 250`)
- [x] Checkpoint discovery validated — both full merged models and LoRA adapters supported
- [x] Full eval end-to-end on VM (blocked by GCP L4 stockouts)

## Known issues
- GCP L4 stockouts in us-central1-a intermittently block VM access
- Headless MuJoCo rendering requires system libs (`libegl1-mesa-dev`) and `MUJOCO_GL=egl` + `PYOPENGL_PLATFORM=egl`
- Evaluation path not yet tested end-to-end on real hardware

---


.claude/settings.local.json
SoyGema, line 1: The .claude folder should not be commited

src/odyssey/runners/openvla.py
```diff
@@ -316,3 +371,167 @@ async def run(self, context: TaskContext) -> dict[str, Any]:
...
             ),
         }
+
+
+# ---------------------------------------------------------------------------
+# Inference policy
+# ---------------------------------------------------------------------------
+
+# Default natural-language instructions per Robosuite benchmark.
+_DEFAULT_INSTRUCTIONS: dict[str, str] = {
```
SoyGema, line 381: Unclear if this set of instructions might be generalizable to other policies, therefore it should appear somewhere. At some point I thought to put it on yaml, but having only one word in yaml is clean and having the same set of instructions is convenient

src/odyssey/runners/openvla.py
```diff
@@ -1,5 +1,7 @@
-"""OpenVLA training runner.
+"""OpenVLA training runner and inference policy.
```
SoyGema, line 1: At some point of this PR, there were 2 different scripts, one for training and another one for inference,  There is a pending discussion on improving a little bit folder structure . For now, I think it´s consistent to group models

src/odyssey/runners/robosuite.py
```diff
@@ -157,7 +188,18 @@ async def run(self, context: TaskContext) -> dict[str, Any]:
             step="load_policy",
             step_label=str(checkpoint),
         )
-        policy = self._policy_factory(checkpoint)
+
+        # Policy: use OpenVLA when no custom factory was injected
+        if self._policy_factory is _default_policy_factory:
+            from odyssey.runners.openvla import make_openvla_policy
+
+            policy = make_openvla_policy(
```
SoyGema, line 196: Unclear if this will scale well when we introduce new models , maybe this fucntion shall be more generic

src/odyssey/telemetry/events.py
```diff
@@ -26,3 +29,23 @@ class TaskEventType(str, Enum):
     COMPLETED = "task.completed"
     FAILED = "task.failed"
     CANCELLED = "task.cancelled"
+
+
+class ProgressEvent(BaseModel):
+    """Structured progress event emitted by runners.
+
+    Mirrors lai-trainer's ``ProgressEvent`` schema so both the local CLI
```
SoyGema: Maybe this comment should be removed. Reviewer shall know that we mimic cc

CLAUDE.md
SoyGema, line 1: shall not be commited

tests/unit/test_openvla.py
```diff
@@ -107,18 +107,35 @@ def test_argv_does_not_override_explicit_use_lora(tmp_path: Path) -> None:
     assert argv.count("--use_lora") == 1
 
 
-def test_argv_passes_dataset_via_data_root_dir(tmp_path: Path) -> None:
+def test_argv_passes_dataset_ref_as_data_root_dir(tmp_path: Path) -> None:
     from odyssey.spec import DatasetRef, DatasetSource
     task = _task(
         dataset=DatasetRef(
-            source=DatasetSource.HUGGINGFACE, ref="lerobot/bridge_v2"
+            source=DatasetSource.OXE, ref="bridge_orig"
```
SoyGema, line 114: what happens here when we introduce HF ?

SoyGema (commented):
Autoreview 

Review some spotted things

src/odyssey/runners/openvla.py
```diff
@@ -1,5 +1,7 @@
-"""OpenVLA training runner.
```
SoyGema, line 1: **Addressing review comments:**

**`.claude/` and `CLAUDE.md`** — Fixed in `0915fc5`. Removed from tracking, added to `.gitignore`.

**`_DEFAULT_INSTRUCTIONS`** — Keeping the dict as a reasonable default set for now. User can override via `config.task_instruction` in mission YAML. When we support more policies, these should move to a shared location (benchmark registry or eval task spec).

**Folder structure (training + inference)** — Consolidated in `153ebad`. `openvla_policy.py` removed, all policy code in `openvla.py`. Grouping by model is the right pattern. When more models arrive we can split into `runners/openvla/train.py` + `runners/openvla/policy.py`.

**`make_openvla_policy` scalability in robosuite.py** — Good point. Currently hardcoded as fallback. Should become a policy factory registry keyed by model type (like `RunnerRegistry` dispatches by task kind). Single model path is sufficient for v0.1.0-alpha.

**ProgressEvent docstring** — Removed the internal reference in `0915fc5`.

**Test with HF datasets** — The test uses `DatasetSource.OXE` because OpenVLA needs RLDS format. HF dataset support for other runners would get separate tests. The OXE and HF paths are different code branches and don't conflict.

src/odyssey/telemetry/events.py
SoyGema: solved

SoyGema:
## End-to-end pipeline validated on GCP L4 VM 🎉

**Date:** 2026-06-11

Both tasks in the quickstart mission (`finetune-openvla` → `eval-on-robosuite-lift`) now run end-to-end without errors.

### Training
- `max_steps: 10` for quick validation
- Checkpoint saved successfully

### Evaluation
- `RobosuiteRunner` auto-wires to `make_openvla_policy()` — no custom `policy_factory` needed
- MuJoCo headless rendering works with `MUJOCO_GL=egl` + `PYOPENGL_PLATFORM=egl`
- 10 episodes completed (all FAIL / return=0.0 — expected with only 10 training steps)
- No crashes, no errors

### Fix in this session
- `predict_action` was called with raw `(image, instruction)` but the HF-hosted `OpenVLAForActionPrediction.predict_action()` expects `(input_ids, unnorm_key)`. Fixed by pre-tokenizing with the processor first (commit 4c3087a).

### Environment
```
GPU: NVIDIA L4 (24GB)
MUJOCO_GL=egl
PYOPENGL_PLATFORM=egl
NCCL_NET=Socket
OPENVLA_REPO_PATH=~/openvla
WANDB_MODE=disabled
```

This closes the original issue #5 scope (torchrun + invalid CLI args) **and** validates the eval pipeline that was previously blocked.

SoyGema:
Captura de pantalla 2026-06-11 a las 10 51 58 1

femtechie (dismissed):
Nice work on this change. I was able to get the OpenVLA fine-tuning workflow running successfully with the `bridge_orig` OXE dataset, and the dataset reference resolution appears to be working as intended.

I think this is a meaningful improvement to the developer experience. Being able to reference an OXE dataset directly and have the workflow handle the resolution/loading logic removes a lot of manual setup and makes the fine-tuning workflow much more approachable. This should make it easier for developers to get from "I want to train on this dataset" to an actual training run without needing to understand all of the underlying dataset plumbing.

Most of the feedback below is not about the implementation itself, but about the onboarding experience. During validation, the majority of my time was spent working through environment setup, dependency issues, and understanding where artifacts were being stored. I think a few documentation and usability improvements would make this feature much easier for others to adopt and validate.

### 1. Environment & Dependency Requirements

I ran into several dependency-related issues before training could start, including:

* OpenVLA-specific dependencies (e.g. `draccus`)
* Protobuf/TensorFlow/TensorFlow Metadata version compatibility issues
* W&B configuration requirements

A documented set of expected dependency versions, installation steps, or a known-good environment configuration would make onboarding and validation much smoother.

### 2. Dataset Download Location

This was probably the biggest source of confusion during testing.

The mission configuration references:

```yaml
dataset:
  source: oxe
  ref: bridge_orig
```

but the dataset was actually being downloaded and stored under:

```text
~/bridge_dataset/1.0.0/
```

Since this behavior wasn't documented, it initially looked like the dataset wasn't being downloaded or was being placed in the wrong location. A short note about the expected download location, naming convention, and how dataset references map to the underlying TFDS datasets would save a lot of debugging time.

### 3. Data Root Directory Expectations

The example mission currently contains:

```yaml
data_root_dir: /path/to/dataset
```

It would be helpful to explain:

* What `data_root_dir` should point to
* Where datasets are downloaded by default
* How `bridge_orig` resolves to the underlying dataset
* A known-good example configuration that can be run end-to-end

Having a fully working example would make validation significantly easier for future users.

### 4. W&B Requirements

Training currently stops if W&B isn't configured or if the configured entity/project isn't accessible.

It would be helpful to document:

* How to configure W&B
* How to disable W&B for local testing
* Any expected entity/project settings

This would help users quickly distinguish between a training issue and a logging/authentication issue.

### 5. Progress Visibility / Developer Experience

Not an issue with the implementation itself, but one thing I found myself wondering repeatedly during testing was: *"Is this progressing normally?"* and *"How much longer should I expect this stage to take?"*

There are several stages where users may be waiting for a while:

* Model downloads
* Dataset downloads
* Dataset indexing/loading
* Checkpoint loading
* Training startup

I don't think it's realistic to provide universal time estimates since those will vary significantly based on hardware, storage performance, network bandwidth, and dataset location. That said, it would be helpful to provide some expectation-setting either in the documentation or through logging.

A few examples:

* Expected dataset size
* Expected number of train/validation shards
* Which stage the workflow is currently in
* Rough guidance on what users should expect on a reference environment

Even simple messages such as "expecting 1024 train shards and 128 validation shards" or "downloading Bridge dataset (~X GB)" would help users distinguish between a long-running operation and a potential issue.

For context, I performed validation on a Lambda GPU VM with a single H100. I realize this is not the minimum recommended environment, so I wouldn't expect any timing guidance to be prescriptive. However, including reference numbers from a known-good setup (or documenting the environment used to produce those numbers) would give users a useful baseline and make the overall experience feel much more transparent.

### Overall

The core functionality worked and I was ultimately able to get training running successfully. I appreciate the effort that went into simplifying dataset handling, and I think this will be a valuable quality-of-life improvement for anyone working with OpenVLA fine-tuning workflows.

My main recommendation would be to accompany this change with a README/quickstart update that captures the environment requirements, dataset expectations, and a bit more guidance around long-running operations. Doing so would help future developers benefit from the feature immediately without having to rediscover some of the setup details that came up during validation.

Overall, great work on the feature. I think a relatively small investment in documentation and expectation-setting would make an already useful improvement feel much more polished and approachable for new users.

SoyGema:
Thanks for the thorough validation, @femtechie — and especially for separating the implementation (which worked) from the onboarding friction (which was real). I've addressed all five points in `08474a7`, scoped to documentation only (no behavior change):

**1. Environment & dependency requirements** — Added an *Environment setup (known-good)* section with the pinned stack OpenVLA expects (Python 3.10, torch 2.2.0, transformers 4.40.1, tokenizers 0.19.1, timm 0.9.10, flash-attn 2.5.5, draccus). The key clarification: these deps live in the **cloned OpenVLA repo**, not Odyssey's `openvla` extra — which is exactly why the protobuf / TensorFlow / `tensorflow-metadata` / `draccus` conflicts surface. The README now points to OpenVLA's repo as the source of truth and shows installing its requirements.

**2. Dataset download location** — Documented that **Odyssey does not download the dataset**: `source: oxe` is a pass-through (there's no OXE provider). `ref` → `--dataset_name`, `config.data_root_dir` → `--data_root_dir`, and OpenVLA/TFDS loads from `<data_root_dir>/<folder>/<version>/`. I also called out the naming gotcha you hit explicitly: `ref: bridge_orig` resolved to `~/bridge_dataset/1.0.0/`, so `data_root_dir` must point at the *parent* of that folder, not the key name.

**3. `data_root_dir` expectations** — Explained as the parent dir of the on-disk RLDS folder, and fixed the misleading comment in `examples/quickstart-openvla/mission.yaml` (it claimed `bridge_orig/1.0.0/`, which contradicts the actual `bridge_dataset/1.0.0/` layout).

**4. W&B** — Documented `WANDB_MODE=disabled` for local testing, and `wandb_project` / `wandb_entity` via mission config (forwarded as passthrough `--wandb_*` flags). Noted that `finetune.py` calls `wandb.init()` unconditionally, so this helps separate a training failure from a logging/auth failure.

**5. Progress / expectation-setting** — Added a *What to expect during a run* section with stage-by-stage orientation (base model ~14 GB download, Bridge V2 ~124 GB / 1024 train + 128 val shards, RLDS indexing, then steady-state `it/s` with a ~1.49 it/s reference on an L4), framed as orientation rather than prescriptive timing — same point you made about hardware variance.

One honesty note: the dependency pins and dataset-download mechanics come from OpenVLA's own docs (I can't exercise the full OpenVLA stack from here), so the README defers to OpenVLA's repo and links out rather than asserting its internals. If anything in the *Environment setup* block doesn't match what you actually ran on the H100, that's the most likely spot for a correction — happy to adjust to your validated numbers.

femtechie (approved):
I confirm the doc changes are in. I think those will be very helpful. It's a good idea that we are  keeping a clean boundary with the model repo as source of truth. Let's commit!

PR #24, opened 2026-06-18
feat(eval): GR00T policy eval recipe for the Isaac Lab runner (#17)

SoyGema:
## What
Brings the **blessed GR00T evaluation recipe** onto `develop`. This is the content of **#23**, which got merged into the now-superseded `gr00t-quickstart-runner` branch (after #18 had already squash-landed on `develop`), leaving the recipe orphaned — never reaching `develop`. This PR re-applies exactly those 7 files on top of `develop`.

Fills the `eval_script` slot the subprocess `IsaacLabRunner` leaves open: the runner owns launch / the `ODYSSEY_*` stdout protocol / cancellation / scoring; this recipe owns the GR00T↔Isaac obs/action loop — boots the Franka visuomotor env, drives a GR00T policy server, and reports each episode.

## Changes
- `scripts/eval/gr00t_isaac_eval.py` — the eval recipe
- `scripts/eval/gr00t_transforms.py` — quaternion / rot6d / matrix helpers
- `examples/quickstart-gr00t/isaac_eval_mission.yaml` — eval mission
- `examples/quickstart-gr00t/mission.yaml` — wire the eval task
- `tests/unit/test_gr00t_isaac_eval.py`, `tests/unit/test_gr00t_transforms.py`
- `tests/unit/test_spec.py` — spec assertions for the wired eval task

## Notes
- Original PR #23 (merged into `gr00t-quickstart-runner`, **not** `develop`) is superseded by this.
- Minor lint cleanup to satisfy the repo's ruff gate: removed dead `noqa` directives + an unused import, split a semicolon line.

## Verification
- `ruff check` clean · `mypy --strict` clean (54 files) · **244 passed, 4 skipped** (conformance skips need an Isaac-GR00T checkout)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #23, opened 2026-06-17
feat(eval): GR00T policy eval recipe for the Isaac Lab runner (odyssey#17)

Danny024:
## What
Fills the `eval_script` slot the subprocess `IsaacLabRunner` leaves open: the **blessed GR00T evaluation recipe**. The runner owns launch / the `ODYSSEY_*` stdout protocol / cancellation / scoring; this recipe owns the GR00T↔Isaac obs/action loop — boots the Franka visuomotor env, drives a GR00T policy server, and reports each episode.

## Changes
- `scripts/eval/gr00t_isaac_eval.py` — the recipe. Accepts the launch contract (`--task/--num_episodes/--checkpoint/--headless` + config passthrough), connects to `run_gr00t_server`, replays GR00T action chunks in the IK-rel env, emits `ODYSSEY_EPISODE`/`ODYSSEY_RESULT`. Heavy deps (isaaclab/gymnasium/gr00t/torch/numpy) are deferred into the run path so the argv + protocol surface unit-test on a CPU box without Isaac. Optional `--video_dir` captures Isaac's viewport render (3rd-person, 720p).
- `scripts/eval/gr00t_transforms.py` — pure-numpy obs/action transforms (rot6d/quat/axis-angle; nested obs build; action decode).
- `tests/unit/test_gr00t_isaac_eval.py` — parser + `ODYSSEY_*` lines verified through the runner's own `EvalProtocolCollector` + `summarize`.
- `tests/unit/test_gr00t_transforms.py` — transform conventions (numpy-gated).
- `examples/quickstart-gr00t/isaac_eval_mission.yaml` — runnable mission wiring the recipe.

## Validation
Ran live on an NVIDIA H100 (GR00T server on GPU + Isaac Lab cosmos env): episode completed, `success_rate 0.0` (expected zero-shot for base GR00T) — the GR00T-in-Isaac loop is proven. Also driven end-to-end from Command Center via the CC→Odyssey delegation bridge.

PR #16, opened 2026-06-11
feat: multi-agent evaluation pipeline (OpenVLA + Gemma + Robosuite)

SoyGema:
## Summary

Multi-agent **evaluation** support: a **PILOT** (OpenVLA 7B) guided by a **SPECIALIST** (Gemma 2B int4 task planner), evaluated on **Robosuite** — plus a reorganization of the `runners/` package into `models/`, `evals/`, `agents/`.

> **Scope note:** Isaac Lab was removed from this PR. #16 and #18 both add an `IsaacLabRunner` for `evaluation_type: isaac_lab` and can't merge together; #18's subprocess design (Isaac Sim's bundled Python + stdout protocol) is the robust, NVIDIA-aligned approach, so **Isaac Lab is owned by #18**. The multi-agent orchestration here (`PlannedEvalRuntime`) is simulator-agnostic and can layer on #18's transport later (see the discussion thread). The `EvaluationType.ISAAC_LAB` enum stays as shared vocabulary.

## How PILOT and SPECIALIST interact

During evaluation, the two agents collaborate in a **plan-then-execute** loop:

1. **SPECIALIST plans once per episode** — decomposes the task instruction into ordered sub-steps (`"pick up the red cube"` → `["move above cube", "lower gripper", "close gripper", "lift"]`).
2. **PILOT executes each phase** — receives the current sub-instruction + camera image every sim step, outputs a 7-DoF action. After `steps_per_phase` (default 50) it advances.
3. **The simulator judges** — episode ends on done/truncated; success reported back.

| | SPECIALIST (Gemma 2B) | PILOT (OpenVLA 7B) |
|---|---|---|
| **Role** | Task planner | Action executor |
| **Input** | High-level instruction (text) | RGB image + sub-instruction |
| **Output** | List of sub-steps (text) | 7-DoF action (ndarray) |
| **When** | Once per episode | Every simulator step |
| **Trained** | No (base model, int4) | Yes (LoRA fine-tune) |

**Backward compatible:** missions with only a PILOT use the single-instruction policy path — no behavior change.

## What's in this PR

- **Spec** — `RobotSpec.agents` open to multiple agents, enforce ≥1 PILOT, block SPECIALIST training, `HFModelRef.quantization`.
- **Engine** — `_execute_task` resolves all agents + checkpoints for eval tasks into `TaskContext`.
[...]

PR #18, opened 2026-06-11
feat: Add GR00T quickstart mission, training runner, and Isaac Lab eval runner

Danny024:
Implements #17 — the GR00T path for the June 22 launch and the NVIDIA Automate session. Two commits: the GR00T training side, then the Isaac Lab evaluation side.

## Summary

- **`examples/quickstart-gr00t/mission.yaml`** — fine-tune `nvidia/GR00T-N1.7-3B` (demonstration training) on the LeRobot-v2-flavor demo set shipped inside the [Isaac-GR00T repo](https://github.com/NVIDIA/Isaac-GR00T) (`demo_data/cube_to_bowl_5`, no separate download), evaluated on Isaac Lab `Isaac-Lift-Cube-Franka-v0`. Validates and mock-runs everywhere today.
- **`src/odyssey/runners/gr00t.py`** — subprocess runner invoking `python -m gr00t.experiment.launch_finetune`, mirroring the OpenVLA runner: base-model resolution ladder (task config → `NVIDIA_GR00T_N1_7_3B_PATH`-style env var → HF id) with starting-checkpoint chaining and HF provider fetch; mission config keys pass through 1:1 as kebab-case upstream flags; HF-Trainer stdout parser. Contains no NVIDIA code — users install Isaac-GR00T and accept the upstream weight license themselves.
- **`src/odyssey/runners/isaac_lab.py`** — subprocess eval runner (Isaac Lab scripts must run under Isaac Sim's bundled Python, so in-process like Robosuite isn't an option). Launches a user-supplied eval script via `${ISAACLAB_PATH}/isaaclab.sh -p <script> --task <benchmark> --num_episodes <N> --checkpoint <path> --headless` and scores the `ODYSSEY_EPISODE` / `ODYSSEY_RESULT` JSON lines it prints (boot noise ignored, episodes stream as live progress). The eval script is the deliberate slot for the blessed GR00T/VLA recipe (#17); checkpoint resolution reuses the Robosuite loadout walk so fine-tune → eval composes in one mission. `TrainingProcessSpec` gains an optional `launcher` prefix to support non-python launchers.
- **Task-level runner selection** — OpenVLA and GR00T are both wildcard training runners and the registry can't see the model family, so the engine now honors `config: {runner: <name>}` per task via the registry's previously-unused `override` parameter. `--use-mock-runner` becomes an engine-level `force_runner` that beats any task override, preserving its semantics exactly.

## Review notes

[...]

PR #25, opened 2026-06-18
feat(eval): closed-loop GR00T policy eval — relocate recipe to runners/evals + auto-serve the finetuned checkpoint

Danny024:
## What
Closed-loop GR00T policy eval through Odyssey: a single `odyssey run` does **finetune → deploy the trained checkpoint → grade in Isaac Lab**.

Two pieces:
1. **Relocates** the GR00T eval recipe from `scripts/eval/` into the framework (`src/odyssey/runners/evals/`), beside the `isaac_lab` runner it serves — the top-level `scripts/` placement raised on #23/#24. Sibling `gr00t_transforms` import made robust (script-launch *or* package import).
2. **Adds closed-loop auto-serve** (`--serve_checkpoint true`): the recipe boots a GR00T policy server on the upstream training task's checkpoint, waits for it to be ready, runs the eval against it, then tears it down. #24's recipe connects to an *externally*-started base-model server; this serves the **just-finetuned** checkpoint, closing the train→deploy→grade loop.

## Changes
- `src/odyssey/runners/evals/gr00t_isaac_eval.py` — relocated + `build_server_command`, `serve_checkpoint`, `_wait_for_server`, and `--serve_checkpoint/--embodiment_tag/--server_python/--server_device/...` flags
- `src/odyssey/runners/evals/gr00t_transforms.py` — relocated
- `examples/quickstart-gr00t/closed_loop_mission.yaml` — finetune → auto-serve → eval
- `tests/unit/test_gr00t_isaac_eval.py`, `tests/unit/test_gr00t_transforms.py` — relocation import fixes + 7 auto-serve tests

## Validation (H100)
- 20 recipe/transforms unit tests pass; **303 full suite green** (no regressions)
- `build_server_command` matches the real `run_gr00t_server --help` flags (raw nested obs → no `--use-sim-policy-wrapper`)
- Plumbing run (`--use-mock-runner`) COMPLETED
- **Live:** auto-serve booted a finetuned 3-shard checkpoint to ready (~14s), `PolicyClient` round-trip, clean teardown

## Notes
- Based on `multiagent_run_NVIDIA` (#16) — which carries the merged GR00T runners under the `runners/{models,evals,agents}` taxonomy — so the diff is just these 5 files. Happy to retarget to `develop` once #16 lands.
- The Isaac Lab Cosmos sim rollout itself is #24's recipe (heavy, `ISAACLAB_PATH`-gated); training and eval embodiment tags must match for a meaningful success rate.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -37,8 +37,10 @@ is installed (`pip install -e` of the checkout, plus
 evaluation runs work once you have Isaac Lab installed
 (`$ISAACLAB_PATH`) and supply an eval script speaking the `ODYSSEY_*`
 stdout protocol via `config: {eval_script: ...}` — see
-`src/odyssey/runners/isaac_lab.py` for the contract. The built-in
-GR00T-policy eval script is the remaining v0.2.x piece.
+`src/odyssey/runners/evals/isaac_lab.py` for the contract. The blessed
+GR00T-policy eval recipe now ships at
+`src/odyssey/runners/evals/gr00t_isaac_eval.py` — see **Closed-loop GR00T
+eval** below.
 
 The moving parts:
 
@@ -61,3 +63,42 @@ The moving parts:
   under `isaaclab.sh -p`, passes
   `--task/--num_episodes/--checkpoint/--headless`, and scores the
   `ODYSSEY_EPISODE` / `ODYSSEY_RESULT` lines the script prints.
+
+## Closed-loop GR00T eval (train → deploy → grade in one run)
+
+`quickstart-gr00t/closed_loop_mission.yaml` runs the whole loop in a single
+`odyssey run`:
+
+1. **Train** — a `runner: gr00t` training task fine-tunes GR00T N1.7 on the
+   demo set, producing a checkpoint.
+2. **Deploy + grade** — the `isaac_lab` eval task uses the blessed GR00T eval
+   recipe at **`src/odyssey/runners/evals/gr00t_isaac_eval.py`** (relocated from
+   the top-level `scripts/` folder, so it lives beside the `isaac_lab` runner it
+   serves; `gr00t_transforms.py` is its sibling). With `serve_checkpoint: true`
+   the recipe **auto-serves the just-trained checkpoint** as a GR00T policy
+   server, waits for it to be ready, drives the Franka visuomotor env in Isaac
+   Lab, and reports per-episode `ODYSSEY_*`. The runner auto-passes the training
+   task's checkpoint, so the deployed policy is exactly what was just trained —
+   no manually-started external server.
+
+The recipe keeps its heavy deps (isaaclab / gr00t / torch / numpy) lazy in the
+run path, so its argv + `ODYSSEY_*` surface stay unit-testable on a CPU box.
+Key eval-task `config:` keys:
+
+| key | meaning |
+|---|---|
+| `eval_script` | the relocated recipe path (`src/odyssey/runners/evals/gr00t_isaac_eval.py`) |
+| `serve_checkpoint: true` | boot a GR00T server on the trained checkpoint here, vs. connecting to an external one |
+| `embodiment_tag` | **must match** the training `embodiment_tag` — the server can't infer it |
+| `server_python` | interpreter that has `gr00t` installed (the GR00T venv) |
+| `served_model_path` | optional explicit checkpoint to serve (e.g. a Command-Center-delegated eval whose runner `--checkpoint` is a stub) |
+| `pos_scale` | scales the raw GR00T eef deltas into the IK-rel action |
+
+Plumbing-only (no Isaac / GR00T):
+
+```bash
+odyssey run examples/quickstart-gr00t/closed_loop_mission.yaml --use-mock-runner
+```
+
+A real run needs the GR00T venv plus an Isaac Lab install with the Cosmos
+visuomotor env.
--- a/examples/quickstart-gr00t/closed_loop_mission.yaml
+++ b/examples/quickstart-gr00t/closed_loop_mission.yaml
@@ -1,0 +1,90 @@
+# GR00T closed-loop mission: finetune -> deploy the trained checkpoint -> grade.
+#
+# One `odyssey run` drives the whole loop:
+#   1. training task  (runner: gr00t)      -> finetunes GR00T N1.7 on the demo
+#                                              set; produces a checkpoint.
+#   2. eval task      (evaluation: isaac_lab) -> the relocated GR00T eval recipe
+#        (src/odyssey/runners/evals/gr00t_isaac_eval.py) AUTO-SERVES that
+#        checkpoint (serve_checkpoint: true), waits for the server, runs the
+#        Franka visuomotor env in Isaac Lab against the *finetuned* policy, and
+#        reports success over the ODYSSEY_* protocol — then tears the server down.
+#
+# The runner auto-passes the training task's output checkpoint to the eval as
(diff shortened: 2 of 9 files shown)


SoyGema:
Hi @Danny024 — first, thank you for this. The relocation to `runners/evals/` and the closed-loop auto-serve are exactly the direction we want. 🙌 The recipe genuinely belongs beside the runner, and finetune → deploy → grade in a single `odyssey run` is a great milestone.

To keep things simple, I'd like us to converge on a single source of truth — **branch #16** — and align everything to it:

1. **Let's get #16 reviewed and merged first.** It's the foundation: it carries the `runners/{models,evals,agents}` taxonomy and already has GR00T (#18) integrated. Once it lands on `develop`, the correct `runners/evals/` home exists for everyone, and the remaining PRs can align onto it cleanly instead of racing each other. I've marked **#16 as P0** so it's clearly the one to prioritize — a review there first will balance the rest of the merges.

2. **Then #24 and #25 align onto that.** #24 (the recipe) rebases onto `develop`-with-#16 and lands in `runners/evals/`; #25 rebases off `develop` and keeps just the auto-serve delta. Nothing ships twice, and everything sits in the location you already chose. (#24 and #25 are P1.)

One small workflow thought for next time, shared with all kindness 🙂: when you're *evolving the same feature* — here, the eval recipe — you usually don't need a second branch. #25 is really "the next step of #24," so it could have been a few more commits on #24's branch (or a PR stacked directly on it) rather than a parallel branch that recreates the same files in a new place. That's what left the recipe living in two homes. Same great work — just one lane keeps us out of exactly this kind of reconciliation.

Net: **review #16 first to balance the merges**, then we align #24 and #25 to it. Thanks again — this is solid work and the auto-serve is a real step forward. 🚀

Danny024:
@SoyGema both sides of the GR00T closed loop are ready for review 🎯

**Odyssey #25 (this PR) — ready:**
- Rebased on `develop` (post #16 merge), **CI green on 3.10 / 3.11 / 3.12**, **MERGEABLE**.
- Carries the eval recipe **relocated to `runners/evals`** (the placement you flagged) — so it brings the recipe to `develop` in the right home — plus the **closed-loop auto-serve** (`--serve_checkpoint` boots a GR00T server on the just-fine-tuned checkpoint) and `--served_model_path`.
- 21 recipe contract tests pass.
- **Supersedes #24** — once #25 lands, #24 can close.

**Command Center side — `lovellai-corp/lai-trainer#6` — also ready:**
- P1 (CC drives the GR00T finetune) + P2 (CC auto-serves the just-trained checkpoint for the eval). Clean **+502/−3** diff, MERGEABLE (stacked on #4).
- Together with #25 this closes **train → deploy → grade through Command Center**.

Validated live on the H100: GR00T matched-domain LIBERO **~93%** (base = non-functional, so a clean before/after), **auto-serve ready in ~14 s**, and a full fine-tune → eval → grade mission COMPLETED. Ready whenever you want to review/merge. 🙂

SoyGema (changes requested):
Hi @Danny024 — thank you for this, it's genuinely strong work. 🙌 The recipe is well structured: heavy deps (isaaclab, gr00t, torch, numpy) are
  imported lazily in the run path so the argv + ODYSSEY_* protocol surface stays unit-testable on a CPU box, the dual-import trick in _transforms() is
  a clean solution to the script-vs-package launch, and the quaternion / rot6d math checks out. The closed-loop auto-serve (train → deploy → grade in
  one odyssey run) is a real milestone. A few suggestions below to make it even more robust — all optional, none blocking. 

Can you please add an explanation to this README.md ? https://github.com/lovellai-dev/odyssey/blob/develop/examples/README.md 
What I understand is that you designed another mission and you put the recipy there.

Just 3 things for the review, we are almost there!!!!
  1. `num_envs > 1 `silently scores only one env. -`-num_envs `is parsed and passed to `gym.make(num_envs=...),` but the rollout reads only env 0
  (_np(reward).reshape(-1)[0]) and steps a single action (.unsqueeze(0)). With num_envs > 1 it would run but only grade one environment, giving
  misleading numbers. Since the missions default to 1 the real-world risk is low, but it's an easy footgun — could we either clamp to 1 with a
  log.warning, or raise if >1?

  2. One episode error aborts the whole sweep. If client.get_action() or env.step() raises mid-episode, the exception propagates and the finally
  closes the env — so the remaining episodes and the ODYSSEY_RESULT line are lost. Wrapping each episode body in a try/except (log it, emit
  episode_line(success=False), continue) would make grading much more resilient to a single flaky query.

  3. Gripper polarity is worth a sanity check. In gr00t_action_to_isaac, grip < 0.5 → open (1.0), else close (-1.0). If GR00T's gripper_position
  convention is inverted, the gripper would do the opposite and stacks would fail silently. Could we confirm it against the GR00T server's convention
  and drop a one-line comment justifying the direction?

Danny024:
@SoyGema all addressed + **validated end-to-end** on the Cosmos Franka env (not just unit tests). Thank you — these made the recipe genuinely more robust. 🙏

**1. README** — added a **"Closed-loop GR00T eval"** section to `examples/README.md`: explains `closed_loop_mission.yaml` (train → auto-serve the trained checkpoint → grade in one `odyssey run`), the relocated recipe at `runners/evals/gr00t_isaac_eval.py` (+ its `gr00t_transforms.py` sibling), and a config-key table (`serve_checkpoint`/`embodiment_tag`/`server_python`/`served_model_path`/`pos_scale`). Fixed the stale `runners/isaac_lab.py` path ref too.

**2. `num_envs > 1`** — clamped to 1 with a `log.warning`. E2E-confirmed firing: `WARNING num_envs=2 ignored — single-env (grades env 0 only); clamping to 1.`

**3. One episode aborting the sweep** — each episode body is now wrapped in `try/except`. E2E-confirmed: the run hit a real mid-rollout error (`Server error: Video key 'image' must be in observation`), and instead of crashing, it logged it, emitted `episode_line(success=False)`, and **continued** — exactly the resilience you flagged.

**4. Gripper polarity** — justifying comment added (low `gripper_position` ⇒ env gripper OPEN/+1, matching Isaac's IK-rel convention), and it's **configurable** via the existing `gripper_open`/`gripper_close` kwargs if a checkpoint's convention is inverted. Honest caveat: I flagged it in-code to confirm the direction against the GR00T server on the first *matched* live rollout (the run above hit an obs mismatch before reaching gripper actions). If you have a known-good direction, I'll bake it in.

**Bonus the e2e caught:** `ODYSSEY_RESULT` was being emitted *after* the `finally` that calls `simulation_app.close()` — which ends the process, so the result line was silently lost on real runs (the runner had been falling back to computing it from `ODYSSEY_EPISODE` lines). Moved the emit before sim teardown — `ODYSSEY_RESULT` now prints. 

**Validation:** 15 recipe contract tests green; full e2e on `Isaac-Stack-Cube-Franka-IK-Rel-Visuomotor-Cosmos-v0` (server up, env booted, episode + `ODYSSEY_RESULT` emitted, `RECIPE_EXIT 0`). Pushed in `f59cecc` + `b71301f`. Almost there — thanks again! 🚀

Danny024:
Pushed `2926303` — the red CI was a dead `# noqa: BLE001` that RUF100 flags under ruff 0.15.18 (ruff runs before the tests, so they never got to execute). Verified locally with the CI toolchain: ruff/mypy clean, 304 passed. All three checks are green now, and your earlier requests are in `f59cecc`. Ready for re-review when you have a moment. 🙏

SoyGema (approved):

PR #27, opened 2026-06-19
feat(eval): Cosmos-Reason intent-trace sidecar + eval-protocol consumer

Danny024:
# feat(eval): Cosmos-Reason intent-trace sidecar + eval-protocol consumer

> **PR 1 of 2** for the Tier-3 "reasoning + acting" feature. This is the *mechanism*
> — how a trace is generated and how it's carried through the eval protocol. It is
> **independent of #25** and mergeable now. PR 2 wires the producer (emitting the trace
> from the GR00T eval recipe) and depends on #25.

## Summary

GR00T-N1.7 runs on the **Cosmos-Reason2-2B** backbone (a generative Qwen3-VL). This PR
adds the ability to query that *same* backbone as a VLM and carry a short
natural-language **intent trace** through Odyssey's `ODYSSEY_*` eval protocol — the
"reasoning **and** acting" surface for the demo and Episode Review.

Two decoupled, additive pieces:

1. **Sidecar** — `runners/evals/cosmos_reason.py`
   `ReasoningSidecar.reason(frame, instruction) -> str`: lazy-loads
   `nvidia/Cosmos-Reason2-2B` once, **guarded** so a flaky trace can never fail or stall
   an eval (returns `""` on any error). Pure, stdlib-only protocol surface
   (`build_reasoning_prompt` / `clean_reasoning` / `reasoning_line`) + a standalone CLI
   for narrating over a captured frame. Heavy deps (torch/transformers/PIL/numpy) are
   deferred to the generate path; imported by path so the surface stays unit-testable on
   a CPU box.

2. **Consumer** — `runners/evals/isaac_lab.py`
   `EvalProtocolCollector` parses the optional `ODYSSEY_REASONING` line (distinct from
   `ODYSSEY_RESULT` despite the shared `ODYSSEY_RE` prefix) and `summarize()` carries the
   per-episode traces into `metrics["reasoning"]` so Command Center / Episode Review can
   show reasoning beside the grade.

**Nothing emits the line yet** — that's PR 2. Until then this is inert: an eval that
never emits `ODYSSEY_REASONING` behaves exactly as before.

## Protocol

```
ODYSSEY_REASONING {"episode": 1, "instruction": "...", "reasoning": "..."}
```

Mirrors the existing `episode_line` / `result_line` helpers.

## Try the sidecar (standalone)

The sidecar ships a CLI, so you can narrate over any captured frame without a running
eval — useful for the demo and for sanity-checking the model:

```bash
# in an env with transformers + torch + the cached model:
python src/odyssey/runners/evals/cosmos_reason.py \
  --image frame.png \
  --instruction "stack the red cube on the green cube"
# -> REASONING: Locate the red cube ... grasp ... place it on the green cube.
# -> ODYSSEY_REASONING {"episode": 1, "instruction": "...", "reasoning": "..."}
```

(Emitting the trace *during a live eval* is PR 2.)

## Honest framing

The action policy is what drives the robot. This trace is a *parallel intent narration*
from the same backbone family — a per-episode, scene-grounded plan for human review and
demos, **not** a literal action-by-action rationale and not a control signal.

## Testing (TDD)

- **Unit** — 12 sidecar tests (prompt / clean / protocol-line / lazy-load / guarded /
  frame normalisation) + isaac_lab eval tests extended to 23 (collector records & skips
  malformed reasoning, RESULT-vs-REASONING disambiguation, summary includes-when-present /
  omits-when-absent, and the end-to-end fake-script test emits `ODYSSEY_REASONING` and
  asserts one trace per episode survives the real subprocess → collector → summary path).
  `ruff` + `mypy --strict` clean (toolchain matched to CI: ruff 0.15.18 / mypy 2.1.0).

- **Live (H100)** — the sidecar was exercised against a **real GR00T LIBERO rollout**
  frame and produced a coherent, scene-grounded trace; a perception-only probe (no task
  given) named the actual scene objects (cookie box, stove, 3-drawer cabinet), confirming
  it reads pixels rather than echoing the instruction. Demo panel:
  `demo_assets/demo_panel.png`.

## Live validation caught two real bugs (fixed + regression-tested here)

Both silently produced empty traces and were invisible to the standalone tests/CLI:

1. `device_map=` in `load()` requires `accelerate`, absent in the sim eval venvs →
   load + `.to(device)` (portable to any torch+transformers env).
2. `hasattr(frame, "size")` can't distinguish a numpy array from a PIL image, and sim
   agentviews are flipped (`[::-1, ::-1]`) into negative-stride views transformers'
   fast image processor rejects → new `_to_pil()` (`isinstance` + `np.ascontiguousarray`),
   with unit tests for the negative-stride and PIL-passthrough cases.

## Follow-up

- **PR 2** — emit `ODYSSEY_REASONING` per episode from the GR00T eval recipe
  (`gr00t_isaac_eval.py`); depends on **#25** and on this PR.
- Surfacing `metrics["reasoning"]` in the Command Center / Episode Review UI.

## Checklist

- [x] TDD: tests accompany the change; 12 sidecar + 23 isaac_lab green
- [x] `ruff check src tests` + `mypy --strict` clean
- [x] Additive & guarded — no behavior change when no trace is emitted
- [x] Targets `develop`

### DIFF
--- a/src/odyssey/runners/evals/cosmos_reason.py
+++ b/src/odyssey/runners/evals/cosmos_reason.py
@@ -51,8 +51,21 @@ def build_reasoning_prompt(instruction: str) -> str:
 
 
 def clean_reasoning(text: str, *, max_chars: int = 400) -> str:
-    """Collapse whitespace + truncate the raw generation into a one-line trace."""
-    return " ".join((text or "").split()).strip()[:max_chars]
+    """Collapse whitespace and cap the raw generation into a one-line trace.
+
+    ``max_chars`` is a **hard display ceiling, not the expected length**: at the
+    default 70 ``max_new_tokens`` a trace rarely approaches 400 chars, so the
+    truncation path is a safety net against a runaway generation rather than the
+    common case. When it does fire, cut at the last word boundary (not mid-word /
+    mid-grapheme) and append an ellipsis so the trace still reads cleanly.
+    """
+    collapsed = " ".join((text or "").split()).strip()
+    if len(collapsed) <= max_chars:
+        return collapsed
+    # Truncate on the last whitespace within the budget to avoid a dangling
+    # partial word; fall back to a hard cut if there's no space (single long token).
+    cut = collapsed[:max_chars].rsplit(" ", 1)[0] or collapsed[:max_chars]
+    return cut + "…"
 
 
 def _to_pil(frame: Any) -> Any:
@@ -117,6 +130,11 @@ class ReasoningSidecar:
         from transformers import AutoModelForImageTextToText, AutoProcessor
 
         # Gated Cosmos-Reason2 backbone -> load from the local cache offline.
+        # NOTE: this intentionally mutates *process-wide* env. ``setdefault`` only
+        # sets the flag when unset, so it won't override a caller that explicitly
+        # went online, and any other HF load in the same process wants the same
+        # offline behavior for a gated model. (The GR00T action server runs in a
+        # separate process, so there's no cross-talk with it.)
         os.environ.setdefault("HF_HUB_OFFLINE", "1")
         os.environ.setdefault("TRANSFORMERS_OFFLINE", "1")
         log.info("cosmos-reason: loading %s on %s", self.model_id, self.device)
--- a/src/odyssey/runners/evals/cosmos_reason.py
+++ b/src/odyssey/runners/evals/cosmos_reason.py
@@ -55,6 +55,23 @@ def clean_reasoning(text: str, *, max_chars: int = 400) -> str:
     return " ".join((text or "").split()).strip()[:max_chars]
 
 
+def _to_pil(frame: Any) -> Any:
+    """Normalise a frame (PIL image or HxWx3 array) to a contiguous RGB PIL image.
+
+    Two real-frame hazards this guards against (both caught by a live LIBERO
+    rollout): a numpy array also has ``.size`` so it can't be distinguished from a
+    PIL image by attribute, and sim agentviews are often flipped (``[::-1, ::-1]``)
+    into a negative-stride view that transformers' fast image processor rejects —
+    ``np.ascontiguousarray`` makes a positive-stride copy.
+    """
+    import numpy as np
+    from PIL import Image
+
+    if isinstance(frame, Image.Image):
+        return frame
+    return Image.fromarray(np.ascontiguousarray(frame, dtype=np.uint8))
+
+
 def reasoning_line(*, episode: int, instruction: str, reasoning: str) -> str:
     """One ``ODYSSEY_REASONING`` protocol line — mirrors the recipe's
     ``episode_line`` / ``result_line`` so a runner/collector can pick it up and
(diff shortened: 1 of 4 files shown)


SoyGema (commented):
Reviewed in depth — this is excellent, careful work. 🙌 The design is consistent with the rest of `runners/evals/`: stdlib-only protocol surface that stays unit-testable on a CPU box, heavy deps deferred to the generate path, import-by-path, and a clean `ODYSSEY_*` extension. I especially like that it's **genuinely additive** — `test_summary_omits_reasoning_when_absent` proves an eval that never emits the line behaves exactly as before — and that `reason()` is guarded to never break an eval. The comments capturing real LIBERO-rollout hazards (negative-stride agentview, `device_map` needing `accelerate`) are gold for the next person. Approving in spirit; a few small, non-blocking notes.

### 🟡 Minor

**1. `reason()` never honors the `RESULT`/`REASONING` ordering, but the collector relies on prefix order — worth a guard.** In `EvalProtocolCollector.parse`, `ODYSSEY_RESULT ` is checked before `ODYSSEY_REASONING `. They diverge at the char after `ODYSSEY_RE` (`S` vs `A`) so there's no false match today — but the safety rests on the trailing space in `_RESULT_PREFIX`. A one-line test asserting `"ODYSSEY_RESULT ".startswith(...)` invariants, or a comment on why the space matters, would lock it in. (You already note the shared prefix in the docstring, so this is just belt-and-suspenders.)

**2. `clean_reasoning` truncation can split mid-word / mid-UTF8 grapheme.** `[:max_chars]` on a 400-char trace is fine for display, but truncating to a hard char count can leave a dangling partial word. Optional: truncate on the last whitespace before `max_chars`, or append `…`. Cosmetic only.

**3. `max_new_tokens` default differs across surfaces.** `ReasoningSidecar.__init__` defaults to `70`, and the CLI `--max-new-tokens` also defaults to `70` — consistent, good. But `clean_reasoning`'s `max_chars=400` is an independent cap; with 70 tokens you'll rarely hit 400 chars, so the truncation path is effectively untested in practice. Worth a quick note that 400 is a hard ceiling, not the expected length.

**4. Offline env vars are set as a side effect of `load()`.** `os.environ.setdefault("HF_HUB_OFFLINE", "1")` mutates global process env on first load. That's the right behavior for the gated Cosmos-Reason backbone, but since this sidecar may run in the same process as other HF loads (the GR00T server is separate, but still), a one-line comment that this is intentionally process-wide would help. Non-blocking.

### 🔵 Question (not a change request)

### Tests
Coverage is great — prompt grounding, `clean_reasoning` edge cases (empty/None/whitespace/truncate), protocol JSON round-trip + int coercion, lazy construction, the negative-stride `_to_pil` guard, the never-raises contract, and the RESULT-vs-REASONING disambiguation on the consumer side. The heavy VLM path isn't exercised (correctly — same reasoning as the conformance tests).

Net: clean, well-tested, safely additive. None of the above blocks merge. Thanks for the careful framing in the docstrings — the "honest framing" note about this being parallel intent narration rather than literal action rationale is exactly the right level of candor. 🚀

Danny024:
Thanks for the careful, in-depth review @SoyGema 🙏 — folded all four notes into `82b4f69` (non-blocking, but they're genuine improvements):

1. **RESULT/REASONING ordering** — added a comment at the prefix defs on *why* the RESULT-before-REASONING check is safe (the prefixes diverge at `S`/`A`, and the trailing space prevents prefix-of-longer-token matches), and locked the invariant with **`test_prefixes_are_mutually_exclusive`** (asserts no `ODYSSEY_*` prefix is a prefix of another — guards a future rename/reorder from silently breaking disambiguation).
2. **Mid-word / mid-grapheme truncation** — `clean_reasoning` now cuts at the last word boundary and appends an ellipsis; no dangling partial word (new test covers it).
3. **`max_chars=400` framing** — documented as a *hard display ceiling, not the expected length* (at 70 `max_new_tokens` the truncation path is a runaway-generation safety net); added an under-limit passthrough test.
4. **Process-wide offline env** — added the one-line note that the `setdefault` mutation is intentional and process-wide (only sets when unset, harmless to other in-process HF loads).

On the 🔵 tests question — agreed, leaving the heavy VLM path unexercised on purpose (GPU + gated model; same reasoning as the conformance tests).

✅ 37 unit tests green, ruff 0.15.18 clean, mypy strict (`src`) clean.

Danny024:
@SoyGema — all four of your notes are folded in now (`82b4f69`: word-boundary truncation, the prefix-invariant test, and the two doc comments), **37 unit tests green / ruff + mypy clean**. Your in-depth review came through as a `Comment`, so the PR is still on `REVIEW_REQUIRED`.

When you have a sec, could you convert it to a formal ✅ **Approve**? You have write access here, so your approval clears the gate and this can merge. Thanks again for the thorough pass! 🙏

SoyGema (approved):

PR #31, opened 2026-06-20
docs(readme): remove duplicate intro + Concepts section

SoyGema:
## What
Trims accumulated duplication in the top-level README (built up across successive squash merges).

Removes:
- the **first `## What it is`** (the "You describe a mission…" copy) — the intro reappears later in the doc, so this drops the redundant first copy
- the entire **`## Concepts`** section (Missions · Robots and agents · Robot specs in v0.0.x)

**647 → 507 lines.** No internal links pointed at the removed sections.

## Notes
- Opened as a **draft** against `main` for review.
- ⚠️ The README still has other duplicated subsections further down (Dataset / W&B / "What to expect" appear twice — once flat, once under *Launching a training mission > OpenVLA*). Left **as-is for now** per scope; can be a follow-up.
- `main` and `develop` currently share this README, so a matching trim on `develop` will be needed to keep them aligned (otherwise it returns on the next release).


Danny024 (approved):

femtechie (approved):
Solid PR, thanks for catching the duplicate information! I think we just need to move "What is it?" above Install section and we're good to go. Thanks, Jeanine

PR #36, opened 2026-06-24
fix(eval): stop closed-loop GR00T-Isaac eval from deadlocking on an orphaned server

Danny024:
## Problem
The closed-loop GR00T-in-Isaac eval (`serve_checkpoint: true`) **deadlocks**: on a live H100 run it booted Isaac Sim, the GR00T server loaded the checkpoint, then sat at **0% GPU for ~40 minutes** until the caller's timeout — no `result_summary` ever delivered.

## Root cause
`serve_checkpoint()` spawned the GR00T policy server with `subprocess.Popen(argv, env=env)` — **no stdout/stderr redirect** — so the server **inherited the recipe's stdout/stderr**, which is the pipe Odyssey's `IsaacLabRunner` (and, upstream, Command Center) read. `run_eval()`'s `finally` calls `simulation_app.close()`, which **hard-exits the recipe**, skipping `serve_checkpoint()`'s server-teardown `finally`. The orphaned server (reparented to PID 1) keeps the pipe's write-end open, so the orchestrator's read **never sees EOF** and `communicate()` blocks forever. `ODYSSEY_RESULT` is emitted before the close, but it's never consumed because the pipe never closes.

## Fix
1. **Redirect the server's stdout/stderr to a log file** (`/tmp/gr00t_server_<port>.log`) so it never holds the orchestrator pipe — this alone makes the read EOF as soon as the recipe exits.
2. **`PR_SET_PDEATHSIG`** via `preexec_fn` so the kernel `SIGKILL`s the server if the recipe hard-exits — no orphaned multi-GB GPU process.

The existing teardown `finally` is kept for the normal-exit path.

## Verification
Applied + re-ran the closed-loop eval live on an H100: it now **COMPLETES in ~60–90 s** (vs. the 40-min hang), the result flows back, the server is reaped, and the GPU is freed. (Grade was F on a 500-step smoke checkpoint — expected; the point was that the train→deploy→grade loop closes.)

## Note
Because GR00T (torch 2.7.1) and isaacsim (torch 2.7.0) skew, keep running the GR00T server from the Isaac-GR00T venv (`server_python`) rather than co-installing `gr00t` into the isaac-sim interpreter.


SoyGema:
Heads-up @Danny024 — the Python 3.12 failure was the numpy stub issue we fixed in #33 (`Type statement is only supported in Python 3.12 and greater` from numpy's bundled `.pyi`), nothing to do with your change. I pulled `develop` into your branch so it picks up that fix — CI should go green now. Really solid work on the deadlock fix, thank you! 🙌

SoyGema (approved):
Approving — this is a clean, well-scoped fix for a real, observed bug, and the root-cause write-up in the comments is genuinely excellent. Nice work @Danny024 🙌

The two-part fix is sound:
1. Redirecting the server's stdout/stderr to a log file so the orphaned server never holds the orchestrator pipe's write end — this is what actually kills the hang (runner's read finally sees EOF).
2. `PR_SET_PDEATHSIG` → `SIGKILL` so the kernel reaps the server if the recipe hard-exits via `simulation_app.close()` — good belt-and-suspenders against a GPU leak.

CI is green on 3.10/3.11/3.12, and the error message now points at the server log. 👍

**One non-blocking robustness nit** (worth a quick follow-up): `_die_with_parent` calls `ctypes.CDLL("libc.so.6")` *inside* `preexec_fn`, i.e. after `fork()` in the child. This recipe runs under Isaac Sim, which is heavily multi-threaded, and a post-fork `dlopen` takes the dynamic-linker lock — if another thread holds it at fork time the child can deadlock before `exec`. Low probability at steady-state `Popen`, but ironic in a PR whose whole point is killing a hang. Cheap to eliminate by loading libc *before* `Popen` and keeping `preexec_fn` to just the `prctl` call:

```python
libc = ctypes.CDLL("libc.so.6", use_errno=True)   # before the fork
def _die_with_parent():
    with contextlib.suppress(Exception):
        libc.prctl(1, int(signal.SIGKILL), 0, 0, 0)
```

Minor/optional: `int(signal.SIGKILL)` for ctypes safety, and consider writing the server log under the eval's output dir instead of `/tmp` so it lands as an artifact next to the results (and avoids a predictable name in a shared `/tmp`).

None of these block merge — happy to approve as-is and fold the libc tweak into a quick follow-up. Thanks again for chasing this one down! 🚀

PR #37, opened 2026-06-24
fix(ci): set mypy python_version to 3.12 so it parses numpy's PEP 695 stubs

Danny024:
## Problem
The **`Test (Python 3.12)`** CI job fails at the `mypy src/odyssey` step (and has been failing on `develop` itself, independent of any PR):

```
.../python3.12/site-packages/numpy/__init__.pyi:737: error:
  Type statement is only supported in Python 3.12 and greater  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

mypy is configured `python_version = "3.10"`. On the 3.12 runner the installed **numpy (≥2.2) ships stubs that use PEP 695 `type` aliases** (3.12-only syntax); mypy, told to target 3.10, refuses to parse them → exit 2. The 3.10/3.11 jobs pass because their numpy stubs predate PEP 695. (numpy is in mypy's ignore-list, but `follow_imports = "skip"` doesn't stop the stub from being parsed.)

## Fix
Bump the mypy target to `python_version = "3.12"`. The code already runs on 3.10–3.12, and `pytest` still runs per-version, so version-specific *runtime* issues are still caught.

## Verification
Reproduced the exact error and confirmed the fix with `mypy` + `numpy 2.5.0`:
- `mypy --python-version 3.10` on `import numpy` → `numpy/__init__.pyi:737: ... Type statement is only supported in Python 3.12 and greater` (identical to CI)
- `mypy --python-version 3.12` → `Success: no issues found`

Unblocks the 3.12 job on `develop` and every open PR (incl. #36).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -121,7 +121,7 @@ ignore = [
 ]
 
 [tool.mypy]
-python_version = "3.10"
+python_version = "3.12"
 strict = true
 files = ["src/odyssey"]
 warn_unused_configs = true


SoyGema:
Hi @Danny024 — first off, thank you so much for digging into this and putting up the fix. The numpy + PEP 695 stub breakage on the 3.12 CI was a genuinely thorny one to track down, and your effort here is really appreciated. It's an honor to have you working on Odyssey. 🙏

We're going to close this one — but to be clear, **not** because the change is unwanted. Two reasons:

1. **It was already unblocked in #33.** `develop` now sidesteps the numpy 2.5 PEP 695 stub parse error by skipping those bundled stubs (`follow_imports_for_stubs = true`), so the Python 3.12 CI is green again without changing the type-check target.

2. **Minimum-supported-version compatibility.** We intentionally keep `python_version = "3.10"` in mypy. The project supports Python `>=3.10`, and pinning mypy's target to the *floor* (3.10) is exactly what lets it catch code using 3.11/3.12-only features that would break for our 3.10 users. Bumping it to 3.12 would quietly drop that safety net.

So this is purely about preserving the 3.10 compatibility guarantees — the instinct (get the 3.12 CI green) was spot on, and #33 already gets us there. Thank you again, and please keep them coming! 🚀

PR #38, opened 2026-06-24
[FEATURE] capture rollout video as a near-free eval artifact

SoyGema:
## What

Adds **rollout video capture** to the robosuite eval path (incl. the multi-agent
Gemma-planner + OpenVLA-pilot `PlannedEvalRuntime`). Opt-in via the eval task's
`config.capture_video: true` — saves one MP4 per episode.

## Why

The camera-enabled eval env **already renders an RGB frame every step** (the
policy consumes it), so capturing a clip costs only a per-step list append plus
one encode per episode — the encode is offloaded to a thread so it overlaps the
next episode. Beyond being a nice artifact, a rollout MP4 is the hero asset for
showing the framework actually working (Show HN / `[P]` post).

## How

- New shared helper `src/odyssey/runners/evals/video.py` (`save_rollout_video` +
  `to_uint8_frame`). **Best-effort**: a missing/failed encoder logs a warning and
  the eval still completes — a missing video never aborts an evaluation. Reusable
  by any backend that yields RGB frames.
- Opt-in via `config.capture_video` (+ `video_fps` default 24, `video_format`
  default `mp4`). Robosuite frames are flipped vertically (offscreen renderer is
  bottom-to-top).
- Output: `~/.odyssey/runs/<mission_id>/<task_id>/videos/episode_NN_PASS|FAIL.mp4`,
  surfaced in `result_summary.artifacts.videos` and a `videos_saved` progress event.
- `capture_video: true` enabled in the multi-agent and openvla-quickstart example
  missions; documented in the README multi-agent section.

## Dependency

MP4 encoding needs `imageio[ffmpeg]`, added to the `[robosuite]` (and `[all]`)
extra — no extra step for anyone already running eval. `imageio` added to `[dev]`
for the GPU-free GIF tests.

## Test plan

- [x] Unit tests (GPU-free): helper encodes synthetic frames to GIF; runner path
  uses an injected image-shaped fake env asserting `artifacts.videos`. `313 passed`.
- [x] `ruff` + `mypy --strict` clean.
- [x] **Real GPU run**: validated end-to-end on a GCP L4 (`us-west1-a`) via the
  single-agent `quickstart-openvla` mission (train -> eval). `mission.completed`
  with `result_summary.artifacts.videos` populated — two MP4s written
  (`episode_01_FAIL.mp4`, `episode_02_FAIL.mp4`), one per episode, plus the
  `videos_saved` progress event. (Policy grade F is expected — a 10-step LoRA
  smoke; the point was the capture path.)

## Validated on

GCP **L4** (`us-west1-a`), `quickstart-openvla` mission, single-agent OpenVLA
eval path (`make_openvla_policy`). Both episodes rendered and encoded; the
`artifacts.videos` contract and the `videos_saved` event both fire.

A helper for this exact smoke ships at `examples/quickstart-openvla/setup.sh`
(installs the `imageio[ffmpeg]` encoder, generates a fast machine-local mission:
`max_steps=10`, `num_episodes=2`, `capture_video: true`).

Render-backend note: this VM's NVIDIA driver was compute-only (no NVIDIA EGL), so
the eval rendered via **OSMesa** (`MUJOCO_GL=osmesa`, CPU). Capture is backend-
agnostic — it consumes whatever RGB frame the env renders — so GPU-EGL VMs work
the same, faster. Documented in the GCP tutorial troubleshooting.

## Deferred (noted inline)

Episode-level parallelism: episodes are independent, but a single 24GB GPU is
VRAM-bound by one OpenVLA-7B, so it needs vectorized envs + batched inference or
multi-GPU. Out of scope here.


docs/concepts.md
```diff
@@ -3,8 +3,8 @@
 ## Missions
 
 A **mission** is the unit of work in Odyssey: a single, reproducible recipe
-that fine-tunes a model on a dataset and benchmarks the result. You describe
-one in a `mission.yaml`; the framework loads it, drives it through a
+that trains a multi-agent robot brain by fine-tuning its underlying models on a dataset
```
SoyGema, line 6: Here , it has been changed towards including only multiagent. Previously only VLA finetunning. Both shall be included maybe

docs/overview.md
```diff
@@ -1,8 +1,8 @@
 # Odyssey Overview
 
-An **open-source framework for defining, running, and benchmarking robot training missions**. Built by Lovell AI, currently in pre-alpha (v0.0.1).
+An **open-source framework for defining, running, and benchmarking robot training missions**. Built by Lovell AI, currently in alpha (v0.1.0-alpha.1).
 
-You train an agent by describing a mission in YAML — specifying a robot, a model, a dataset, and an evaluation benchmark — and `odyssey run` orchestrates the full lifecycle: load → validate → train → evaluate → persist results.
+You train a robot with a multi-agent brain by describing a mission in YAML — specifying an embodiment, agents and model, tasks with a dataset, and an evaluation benchmark — and `odyssey run` orchestrates the full lifecycle: load → validate → train → evaluate → persist results.
```
SoyGema, line 5: Same as prvious comment, there shall be a way to reconciliate both worlds

examples/multiagent-openvla-gemma/mission.yaml
```diff
@@ -96,6 +96,10 @@ tasks:
     num_episodes: 10
     config:
       unnorm_key: bridge_orig
+      # Record an MP4 per episode of the rollout. Near-free — the eval env
```
SoyGema: I understand the description, but it shall also be the time to make comments smaller specially in the yaml files. A carefull thought might be needed so it´s consistent across all missions

src/odyssey/runners/video.py
```diff
@@ -0,0 +1,63 @@
+"""Rollout video capture — shared across evaluation runners.
```
SoyGema, line 1: Unclear if at this point we shall need an utils folder and put these things here. Requires a carefull thought

tests/unit/test_rollout_video.py
```diff
@@ -0,0 +1,191 @@
+"""Tests for the rollout-video helper and its wiring into the Robosuite runner.
```
SoyGema, line 1: Tests should be here, definetly

CONTRIBUTING.md
```diff
@@ -4,9 +4,9 @@ Odyssey is an open-source project from Lovell AI. We welcome contributions —
 bug reports, feature ideas, documentation fixes, and patches. This file
 explains what you need to know before sending one.
 
-## Pre-alpha disclaimer
+## Alpha disclaimer
```
SoyGema, line 7: Please check that this inconsistency is also in main,

pyproject.toml
```diff
@@ -12,7 +12,7 @@ license = { text = "Apache-2.0" }
 authors = [{ name = "Lovell AI", email = "odyssey@lovell.ai" }]
 keywords = ["robotics", "machine-learning", "vla", "benchmark", "orchestration"]
 classifiers = [
-    "Development Status :: 2 - Pre-Alpha",
+    "Development Status :: 3 - Alpha",
```
SoyGema, line 15: please check if this happens also in main/dev

.gitignore
```diff
@@ -52,3 +52,8 @@ docs/migration/
 Legacy Mission Service/
 .claude/
 CLAUDE.md
+
+# Security
```
SoyGema, line 56: check sec and video related

README.md
```diff
@@ -331,6 +331,44 @@ python tests/manual/smoke_remote_planner.py
 > (`constraints/specialist-known-good.txt`). They no longer need to be mutually
 > compatible.
 
+### Recording rollout videos
```
SoyGema: this should be at the bottom of the README.md file

SoyGema (commented):
Some minor comments to improve things

docs/concepts.md
SoyGema: Obsolete — this file isn't part of the rollout-video feature; it only appeared in the diff because the branch was based on a stale `develop`. Realigned with current `develop` in 66cf545 (merge), which collapses the PR diff to just the video changes, so this line is no longer in the diff. Resolving.

docs/overview.md
SoyGema: Obsolete — this file isn't part of the rollout-video feature; it only appeared in the diff because the branch was based on a stale `develop`. Realigned with current `develop` in 66cf545 (merge), which collapses the PR diff to just the video changes, so this line is no longer in the diff. Resolving.

CONTRIBUTING.md
SoyGema: Obsolete — this file isn't part of the rollout-video feature; it only appeared in the diff because the branch was based on a stale `develop`. Realigned with current `develop` in 66cf545 (merge), which collapses the PR diff to just the video changes, so this line is no longer in the diff. Resolving.

pyproject.toml
SoyGema: Obsolete — this file isn't part of the rollout-video feature; it only appeared in the diff because the branch was based on a stale `develop`. Realigned with current `develop` in 66cf545 (merge), which collapses the PR diff to just the video changes, so this line is no longer in the diff. Resolving.

.gitignore
SoyGema: Obsolete — this file isn't part of the rollout-video feature; it only appeared in the diff because the branch was based on a stale `develop`. Realigned with current `develop` in 66cf545 (merge), which collapses the PR diff to just the video changes, so this line is no longer in the diff. Resolving.

README.md
SoyGema: Done in 5284d8b — pulled the rollout-video docs out of the Multi-agent `<details>` into a standalone top-level **Recording rollout videos** section near the bottom of the README (and generalized the wording, since capture applies to any Robosuite eval, not just multi-agent). Resolving.

examples/multiagent-openvla-gemma/mission.yaml
SoyGema: Done in 7ea596a — collapsed the 3-line comment to a single inline note (`capture_video: true   # one MP4 per episode (README: "Recording rollout videos")`) in both missions, with the full explanation living in the README. Resolving.

Note: the broader "shorter, consistent comments across *all* missions" cleanup touches files outside this PR's scope (it'd re-pollute the diff), so I'd track that as a separate follow-up rather than fold it in here.

src/odyssey/runners/video.py
SoyGema: Resolved differently than 'utils/': moved to `runners/video.py` in 76e9297. Rationale — `evals/` holds evaluation *environments/backends* (robosuite, isaac_lab, cosmos_reason), whereas this is a cross-runner *utility* that produces an artifact, so it belongs beside the other runner infrastructure (`subprocess.py`, `base.py`), not under evals/ and not in a generic utils/ dumping ground. A broader 'consolidate runner shared helpers' regroup (incl. a future `_common.py` for the eval-results contract) is tracked as a separate follow-up.

SoyGema:
https://github.com/user-attachments/assets/915c2ab4-89da-4f6d-92b9-fdd82ec90d3b

Danny024:
This feature is great and very valid, and important for evaluation. Although I suggest the video frame does not seem to capture the full robot with a good camera angle.

SoyGema:
@Danny024 thank you — that's a sharp observation, and it's exactly right. Here's what's happening and the options I see; I'd like your take before changing anything.

## What's happening today

The captured video frame **is the policy's own observation image** — we deliberately reuse the frame the policy already consumes each step rather than doing a second render (that's the "near-free" rationale of this PR). In `src/odyssey/runners/evals/robosuite.py`:

```python
# single-agent path
if capture_video:
    _capture_frame(frames, _extract_image(obs, image_key))   # ~L319
action = policy(...)
```

And the eval env is built with robosuite's **default `agentview` camera at 256×256** (`_make_eval_env`):

```python
camera_names  = cfg.get("camera_names", "agentview")    # ~L126
camera_height = cfg.get("camera_height", 256)
camera_width  = cfg.get("camera_width", 256)
image_key     = cfg.get("image_key", "agentview_image")  # ~L200
```

So the clip is literally "what the model sees": `agentview`, low-res, framed for policy input — **not** framed to show the whole arm nicely. That's why the robot isn't fully captured.

## The key constraint

Because the recorded frame **is** the policy input, you can't just pick a prettier camera without **also changing what OpenVLA sees** — and the policy expects an agentview-style view, so a different angle could degrade its actions. Camera choice here isn't purely cosmetic today.

## Options (looking for your opinion)

1. **Bump resolution only** (`camera_height`/`camera_width` in the eval `config`, e.g. 512). Same angle, crisper image. Simplest — but it also changes the policy's input size, so not free of side effects.
2. **Switch `camera_names`** to a wider view (`frontview` / `birdview`) via config. Better framing, **but** it changes the policy's input camera → can hurt OpenVLA's behavior. Not recommended as the default.
3. **Decouple the video camera from the policy camera** (my preferred direction). Robosuite can render multiple cameras at once, so the policy keeps consuming `agentview` while the video records a dedicated camera (e.g. `frontview`) that frames the whole robot. Clean separation of "what the policy sees" vs "what looks good on screen", at the cost of one extra render stream per step.

My instinct: keep this PR scoped (it's validated e2e — the MP4 path works) and do **option 3 as a follow-up**, since it's the proper fix and touches the capture design a bit. But I'm open — would you prefer we fold a quick win (option 1/2 via config) into this PR, or track the dedicated-camera approach (option 3) separately?

Thanks again for catching this. 🙏

Danny024:
Great work and options. I would suggest we can go with option 3 which is to track the dedicated-camera approach separately.

Danny024 (approved):

SoyGema:
Agreed, @Danny024 — thanks! Tracked the dedicated-camera approach (option 3) as a separate enhancement: #44. That keeps this PR scoped to the validated near-free capture path, and the better-framed render lands as its own focused change. 🙏

PR #46, opened 2026-06-30
feat(evals): LIBERO eval integration + Franka examples

SoyGema:
> **Draft** — opening early for visibility while the eval path is validated on GPU.

## What this adds

LIBERO eval integration for an OpenVLA-LIBERO checkpoint (eval-only), plus Franka example missions.

**Core**
- `src/odyssey/runners/evals/libero.py` — new LIBERO evaluation runner (~371 lines)
- Wiring: `cli/commands/run.py`, `runners/evals/_common.py`, `spec/tasks.py`, `pyproject.toml`

**Docs**
- `docs/libero-eval-integration.md` — how the LIBERO eval path fits together

**Examples**
- `examples/franka-libero/` — `mission.yaml`, `mission-multiagent.yaml`, `setup.sh`
- `examples/franka-pickplace/` — Robosuite PickPlaceCan mission + README (single- and multi-agent)

## Notes
- Targets `develop` per the repo workflow.
- Eval-only: trains nothing new; runs an existing OpenVLA-LIBERO checkpoint through LIBERO.
- Kept in draft until the eval is validated end-to-end on a GPU VM.


SoyGema:
### Pending validation: multi-agent LIBERO mission + single-vs-multi comparison

So far only the **single-agent** mission (`examples/franka-libero/mission.yaml`) has been run on the VM (GCP L4). The **multi-agent** variant (`examples/franka-libero/mission-multiagent.yaml`, OpenVLA pilot + out-of-process Gemma planner) is **not yet validated**.

- [ ] Run `mission-multiagent.yaml` on the L4 and confirm:
  - [ ] the Gemma specialist loads out-of-process (`env_specialist` + `ODYSSEY_SPECIALIST_PYTHON`),
  - [ ] a per-episode `episode_plan` event is emitted,
  - [ ] both models fit in VRAM (int4 planner ~5 GB + bf16 pilot ~14 GB ≈ 19 GB peak on 24 GB).
- [ ] **Compare single- vs multi-agent** on the same suite/`task_id`: does the planner's phase decomposition change `success_rate` substantially? (LIBERO's longer-horizon tasks are where it's *meant* to help.)

Blocked-ish behind the single-agent `success_rate` investigation in #61 (variance vs center-crop) — worth settling that first so the comparison has a trustworthy baseline.

SoyGema:
## Investigating the ~0 success rate: render backend hypothesis (osmesa vs EGL)

Following up on #61 (finetuned OpenVLA-LIBERO checkpoint approaches the object but fails to grasp, ~0/2). **This is a working hypothesis under investigation, not a confirmed fix** — documenting it here so the reasoning and the test are traceable.

### Hypothesis

VLA policies are highly sensitive to **visual distribution shift**. The two MuJoCo render backends do not produce identical pixels:

- **osmesa** — software render on CPU (Mesa). What we had been running on the headless VM.
- **EGL** — hardware render on the GPU (NVIDIA).

They differ in antialiasing, shading, texture filtering and lighting. The published checkpoint was fine-tuned on GPU-rendered frames, so the hypothesis is that **osmesa observations were out-of-distribution** for the policy, which would explain "arm reaches the object but never closes the grasp cleanly" while our preprocessing (image flip, gripper mapping, center-crop, unnorm_key) already matches OpenVLA's reference `run_libero_eval.py` — i.e. no integration/logic bug was found on those axes.

Framed differently: if EGL resolves it, the root cause was an **environment/render-backend mismatch**, not a code bug in the runner.

### Test being run

On the eval VM (`odyssey-training-west`, L4) we enabled hardware EGL rendering, which previously failed headless:

1. Recovered the NVIDIA kernel driver (kernel/driver module mismatch after a kernel upgrade).
2. Installed the missing NVIDIA GL/EGL userspace runtime (`libnvidia-gl-580-server` → provides `libEGL_nvidia.so` / `libGLX_nvidia.so` + the GLVND vendor JSON). This was the actual reason EGL was unavailable and we had fallen back to osmesa.
3. Re-ran the eval with `MUJOCO_GL=egl` / `PYOPENGL_PLATFORM=egl` (single-agent `mission.yaml`, reduced `num_episodes` / `max_steps_per_episode` for fast iteration).

### How we read the result (diagnostic either way)

- **success_rate > 0 and the rollout video shows a clean grasp** → hypothesis confirmed: the render backend was the cause.
- **Still ~0** → the render backend is not (the whole) cause. Next step is a **bisection** against OpenVLA's own `experiments/robot/libero/run_libero_eval.py` on the same checkpoint + EGL:
  - reference also ~0 → deeper environmental issue;
  - reference 70–90% → residual preprocessing gap in our runner (compare frame-by-frame: RGB channel order, crop interpolation, resize path).

Note: `center_crop` was tried earlier and did not fix it on its own, so we're treating render backend as the leading (but not sole) suspect.

SoyGema:
## UPDATE — the render backend is NOT what fixes the grasp (correcting the hypothesis above)

**TL;DR: the arm grasps the target object in BOTH osmesa and EGL.** So the earlier hypothesis — that osmesa's CPU render caused the grasp failure via distribution shift — does **not** hold as the explanation for the grasp: switching backend is not what changed the behaviour. Correcting the record from the previous comment.

### What we now observe
- In **both** render backends the policy reaches and **grasps** the target object (verified on the rollout videos).
- The episode is still marked **FAIL** because LIBERO success is the BDDL goal — *"pick up X **and place it in the basket**"* — not the grasp alone. In the rollouts the object ends up in the gripper but not yet in the basket.

### New question under investigation: horizon vs policy
The earlier runs had **no explicit `max_steps_per_episode`**, so they ran at the runner default (**1200**) — i.e. the grasp-but-FAIL rollouts were already at 1200 steps, not a shorter budget. We are now re-running with an explicit **2000** to rule out the episode simply running out of steps mid-place:
- **Reaches the basket → PASS** → the residual FAIL was **horizon** (not enough steps to finish the place).
- **Still FAIL with a clean grasp at 2000** → **not horizon**; the checkpoint grasps but does not complete the place → next step is a **bisection** against OpenVLA's own `experiments/robot/libero/run_libero_eval.py` on the same checkpoint + env.

Also being verified: that the grasped object matches the task instruction (`libero_object` has multiple objects on the table; grasping the wrong one can never satisfy the goal, regardless of step budget).

### Note on EGL
EGL is still worth having for **GPU-side render speed** vs osmesa's CPU render, but it is **not** the fix for the grasp — that framing in the previous comment was premature.

Danny024 (approved):

SoyGema:
## Follow-up: multi-agent validation deferred to a separate PR

Following up on the multi-agent validation checklist above — **we're splitting multi-agent out of this PR and tracking it separately.** The reason: we're going to change *how the multi-agent runtime integrates into inference*, so validating the current `mission-multiagent.yaml` wiring now would mean testing a path that's about to change. It doesn't make sense to gate this PR on it.

### What this PR does deliver (single-agent)
- **Single-agent LIBERO eval is validated on the L4** and works end-to-end on `task_0` (`examples/franka-libero/mission.yaml`): the OpenVLA-LIBERO checkpoint loads, runs rollouts, emits events, and produces a `success_rate`.
- The remaining single-agent items (full-suite sweep across `task_id`s, the grasp-vs-place / horizon investigation from #61) are still open, but the **first single-agent integration is considered good** and the eval path is sound.

### Deferred to the multi-agent follow-up
Everything in the checklist above moves to the new PR/issue:
- out-of-process Gemma planner load (`env_specialist` + `ODYSSEY_SPECIALIST_PYTHON`),
- per-episode `episode_plan` event,
- combined-VRAM check (int4 planner + bf16 pilot),
- single- vs multi-agent `success_rate` comparison,

to be re-validated against the **new inference integration** rather than the current wiring.

So: merging this on the single-agent path; multi-agent to be handled on its own once the inference integration lands.

PR #49, opened 2026-07-03
fix(runners): watchdog deadline + reap for the training/eval subprocess

Danny024:
## What & why

`run_training_subprocess` awaited `proc.wait()` with **no timeout**, and
`MissionEngine` runs tasks sequentially with no per-task deadline (the module
docstring literally says "No watchdog timers"). So a hang during **Isaac Sim
boot**, the **ZMQ handshake** to a GR00T policy server that never binds, or a
wedged CUDA context blocks the whole mission **forever** — no `CANCELLED`/`FAILED`
is ever emitted and the GPU stays pinned until a human kills it. This hardens the
training/eval subprocess runner end to end.

(Distinct from the already-fixed closed-loop *server-orphan* deadlock — #36. That
fixed the recipe; this hardens the **runner**.)

## Scope — six findings (F14/F16/F58 + F17/F18/F61 via the merged #51)

**Watchdog + reap (the core):**
- **F14 — deadline.** `TrainingProcessSpec.timeout_seconds`; `run_training_subprocess`
  enforces it via `_wait_with_deadline`: SIGTERM the child's **process group** on
  deadline → wait `sigterm_grace_seconds` → SIGKILL → raise `SubprocessTimeoutError`.
  The three callers (openvla, gr00t, isaac_lab) plumb the task's `timeout_seconds`.
  `None` = unbounded (legacy — no regression risk).
- **F58 — reap-on-crash.** `PR_SET_PDEATHSIG(SIGKILL)` on the child (via
  `_child_preexec`, alongside `setsid`) so a runner crash can't orphan a GPU job.
- **F16/F58 — bounded stdout drain.** The post-exit `await stdout_task` is now
  `asyncio.wait_for`-bounded, so a grandchild that inherited the pipe can't wedge
  the runner waiting for an EOF that never comes.

**Further subprocess hardening (folded in via the merged #51):**
- **F17 — reap-on-any-exit.** `_reap_if_alive` in the `finally` SIGTERM/SIGKILLs
  the child group on *any* exit (a cancelled runner task, an unexpected error),
  not just the deadline path — PDEATHSIG only covers a whole-runner crash.
- **F18 — PYTHONPATH isolation.** `_build_child_env` strips a leaked orchestrator
  `PYTHONPATH` (ROS Jazzy, an editable checkout) from the child env unless the
  spec sets it, so it can't shadow the training env's torch/prismatic.
- **F61 — resource classification.** A bounded output tail is scanned on a
[...]

PR #50, opened 2026-07-03
fix(cli): graceful SIGTERM/SIGINT shutdown of a running mission

Danny024:
## What & why

The mission CLI ran `start_mission` with **no OS signal handlers**. So a **Cloud
Run revision swap** or a **preemptible-VM stop** (SIGTERM), or a **Ctrl-C**
(SIGINT), just killed the process outright: the in-flight GR00T/Isaac subprocess
was **orphaned** (kept holding the GPU) and the mission row was left stuck
**ACTIVE** with no CANCELLED transition. (Finding **F15**.)

## Fix

- **`MissionEngine.request_cancel(mission_id) -> bool`** — a *synchronous,
  signal-safe* method (no await, no I/O) that sets the mission's `cancel_event`.
  `start_mission`'s loop then breaks and `_finalize_mission` performs the
  **CANCELLED** transition (persisted + `CANCELLED` event published), while the
  in-flight subprocess is SIGTERM'd by its existing `_watch_cancel`. Returns
  False if no mission is currently active.
- The CLI's `_run_mission` installs **SIGTERM + SIGINT** handlers that call it,
  echoes a "cancelling…" notice, and **tears the handlers down** in a `finally`.
  A **second signal** reverts to the default disposition (force-quit) for an
  impatient operator.

Complements #49 (which bounds *hangs* with a per-task deadline): together, a
mission can no longer wedge or orphan a GPU job — whether it hangs, or the host
tells it to stop.

## Tests

`tests/test_graceful_shutdown.py` (RED→GREEN): `request_cancel` sets the event
for a running mission, returns False for an unknown one, and is synchronous
(safe from a signal handler); the CLI wires SIGTERM/SIGINT to it and removes the
handlers afterward. Engine unit suites (41 tests) still pass.

Independent of #49 — different files (`mission_engine.py`, `cli/commands/run.py`);
branches off `develop`.

### DIFF
--- a/src/odyssey/cli/commands/run.py
+++ b/src/odyssey/cli/commands/run.py
@@ -15,7 +15,9 @@ network access — useful for smoke-testing the plumbing.
 from __future__ import annotations
 
 import asyncio
+import signal
 import sys
+from contextlib import suppress
 from pathlib import Path
 
 import click
@@ -146,4 +148,35 @@ def run(
 async def _run_mission(engine: MissionEngine, spec: Mission) -> MissionRun:
     await engine.initialize()
     run_record = await engine.create_mission(spec)
-    return await engine.start_mission(run_record.id)
+
+    # Graceful shutdown: on SIGTERM (Cloud Run revision swap / preemptible-VM
+    # stop) or SIGINT (Ctrl-C), cancel the mission so the in-flight GPU
+    # subprocess is SIGTERM'd via its watcher and the run is persisted CANCELLED
+    # — instead of the process dying and orphaning the child with the mission
+    # stuck ACTIVE. A second signal reverts to the default disposition
+    # (force-quit) for an impatient operator.
+    loop = asyncio.get_running_loop()
+    installed: list[int] = []
+
+    def _on_signal(signum: int) -> None:
+        for s in installed:
+            with suppress(Exception):
+                loop.remove_signal_handler(s)
+        installed.clear()
+        if engine.request_cancel(run_record.id):
+            click.echo(
+                click.style(f"\n{signal.Signals(signum).name} received", fg="yellow", bold=True)
+                + " — cancelling after the in-flight task (send again to force-quit)…"
+            )
+
+    for sig in (signal.SIGTERM, signal.SIGINT):
+        with suppress(NotImplementedError, RuntimeError):
+            loop.add_signal_handler(sig, _on_signal, int(sig))
+            installed.append(int(sig))
+
+    try:
+        return await engine.start_mission(run_record.id)
+    finally:
+        for s in installed:
+            with suppress(Exception):
+                loop.remove_signal_handler(s)
--- a/src/odyssey/engine/mission_engine.py
+++ b/src/odyssey/engine/mission_engine.py
@@ -233,6 +233,22 @@ class MissionEngine:
 
         return run
 
+    def request_cancel(self, mission_id: str) -> bool:
+        """Signal a running mission to stop gracefully after the in-flight task.
+
+        Safe to call from an OS signal handler: synchronous, no await, no I/O —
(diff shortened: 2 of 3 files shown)


SoyGema:
Great addition, @Danny024 — thank you! 🙌 The `request_cancel` design is spot on: **synchronous, no await, no I/O, signal-handler-safe**, just setting the `cancel_event` and letting `start_mission` / `_finalize_mission` do the real work off the handler. Installing SIGTERM + SIGINT, echoing a notice, tearing the handlers down in `finally`, and letting a second signal fall through to force-quit — that's exactly the right shape for Cloud Run revision swaps and preemptible-VM stops. Pairs cleanly with #49 (it bounds *hangs*, this handles *the host telling us to stop*). 👏

One thing I'd like to strengthen before merge — **the coverage on the CLI wiring.**

### Why this matters
The failure this PR fixes is *specifically* an integration failure: the process was dying and orphaning a GPU subprocess while the mission row stayed `ACTIVE`. That only happens across the seam between **OS signal → handler → engine → subprocess teardown → persisted terminal state**. A GPU job orphaned on a preemptible stop is expensive and silent — the mission looks alive but nothing is running. So the thing most worth proving is precisely the end-to-end transition, not the individual pieces.

### The gap
`test_cli_installs_sigterm_sigint_handlers_calling_request_cancel` reads `run.py` as **text** and asserts substrings (`"add_signal_handler" in body`, etc.). That's a grep, not a behavioral test — it would stay green even if the handler logic were wired wrong, as long as the keywords appear. The `request_cancel` unit tests are solid; it's the CLI seam that's only string-checked.

### Suggested tests
A couple of behavioral ones would lock this down:

1. **End-to-end cancel → CANCELLED.** Start `start_mission` on a mission whose (mock) runner blocks on an `asyncio.Event`, call `request_cancel(mission_id)` mid-flight, and assert the run finalizes `CANCELLED` (persisted) and a `CANCELLED` event is published. This is the real contract.
2. **Real signal through the installed handler.** Inside a running loop, install the handlers via `_run_mission`'s path and actually deliver `os.kill(os.getpid(), signal.SIGINT)`, then assert `request_cancel` fired. Exercises `add_signal_handler` for real instead of grepping for it.
3. **Handler teardown.** After `_run_mission` returns, assert the loop has no residual SIGTERM/SIGINT handler (no leaked global disposition) — the `finally` cleanup.
4. *(nice-to-have)* **Second signal force-quits.** Assert the first signal removes the handlers so a second one reverts to the default disposition.

(1) and (3) are the high-value ones. Given our GCP setup, we'll also do a manual SIGTERM check on the VM before merge — but a behavioral test means we don't regress it silently later.

Labeled `bug` / `P1-important`. Thanks again — really glad to have this hardening. 🚀

Danny024:
Thanks @SoyGema — great callout, and you're right that a grep isn't a real test. Added behavioral coverage for the seam:

1. **End-to-end cancel → CANCELLED** (the core contract): a mission runs with a `_BlockingRunner` (blocks on `cancel_event`), `request_cancel(mission_id)` mid-flight, and it finalizes **CANCELLED** (persisted) with a `mission.cancelled` event published.
2. **Real signal through the installed handler**: `_run_mission` installs the handlers, then `os.kill(os.getpid(), SIGINT)` is delivered and the run cancels — exercises `add_signal_handler` for real, not by grep.
3. **Handler teardown**: after `_run_mission` returns, `remove_signal_handler` returns `False` for both SIGTERM/SIGINT → no leaked global disposition.

Kept the original `request_cancel` unit tests; the grep test stays as a cheap structural guard. 7 tests, ruff + mypy clean, CI green. Sounds good on the manual SIGTERM check on the VM too — glad to have the behavioral net so it can't silently regress. 🚀

SoyGema (approved):

PR #54, opened 2026-07-06
Guard RemotePlanner cleanup without killpg

### LINKED ISSUE #52: 8 Windows test failures, path separators and os.killpg
Version: 64021721c5b688149239e6f73c5e57d332a570a1
Python: 3.14.3
OS: Windows 10, build 19045.6466
Mission YAML: Not applicable. None of the failures trace to a mission spec.

README lists Linux only for install. This report checks whether the broader test suite runs on Windows regardless, since the base and dev extras installed without a problem. Is Windows support a goal for this project, or should contributors expect to work inside WSL or a Linux VM?

Summary: 324 of 337 tests pass on Windows. 8 fail across two causes, both platform related.

Cause one: path separators.
The GR00T argv builder and the Isaac Lab launcher build paths through string concatenation instead of pathlib or os.path joins. Windows returns backslashes, and the tests expect forward slashes. The same root cause hits the Robosuite runner checkpoint path handling.
Failing tests:
tests/unit/test_gr00t.py::test_argv_resolves_relative_local_dataset_against_repo
tests/unit/test_gr00t.py::test_argv_resolves_relative_modality_config_against_repo
tests/unit/test_isaac_lab.py::test_launcher_from_isaaclab_path
tests/unit/test_robosuite_runner.py::test_runs_all_episodes_and_tallies_successes
tests/unit/test_robosuite_runner.py::test_picks_latest_checkpoint_when_multiple_trainings

Cause two: os.killpg.
RemotePlanner.close calls os.killpg at line 215 in src/odyssey/runners/agents/remote_planner.py. This function exists on Linux and macOS only. Windows raises AttributeError when the line runs, during shutdown of the out of process SPECIALIST planner.
Failing tests:
tests/unit/test_remote_planner.py::test_remote_planner_roundtrip
tests/unit/test_remote_planner.py::test_remote_planner_encodes_image_into_request
tests/unit/test_remote_planner.py::test_remote_planner_falls_back_on_empty_plan

Image

Related gap: tests/unit/test_subprocess.py already skips one test for the same reason, labeled requires POSIX process groups. The skip marks the same gap instead of fixing the gap.

paranoa233:
Addresses the RemotePlanner `os.killpg` cleanup part of #52.

## Summary
- Start the RemotePlanner subprocess with platform-appropriate process creation kwargs.
- Keep POSIX process-group teardown via `killpg/getpgid` when available.
- Fall back to direct `terminate()` / `kill()` when those POSIX APIs are unavailable, avoiding the Windows `AttributeError` during `close()`.
- Add regression coverage for the no-POSIX-process-groups cleanup path.

## Scope note
I dropped the earlier path-separator changes from this PR. The issue discussion clarified that native Windows path support is not currently a project goal, while the RemotePlanner `os.killpg` crash should still degrade cleanly.

## Verification
- `python -m pytest tests/unit/test_remote_planner.py -q` -> 12 passed
- `python -m ruff check src/odyssey/runners/agents/remote_planner.py tests/unit/test_remote_planner.py`
- `python -m mypy --platform linux src/odyssey`

### DIFF
--- a/src/odyssey/runners/agents/remote_planner.py
+++ b/src/odyssey/runners/agents/remote_planner.py
@@ -26,6 +26,7 @@ import os
 import queue
 import signal
 import subprocess
+import sys
 import threading
 import time
 from collections.abc import Sequence
@@ -34,6 +35,24 @@ from typing import IO, Any
 logger = logging.getLogger(__name__)
 
 _SERVER_MODULE = "odyssey.runners.agents.planner_server"
+_SIGKILL = getattr(signal, "SIGKILL", signal.SIGTERM)
+
+
+def _popen_creation_kwargs() -> dict[str, Any]:
+    if sys.platform == "win32":
+        return {"creationflags": subprocess.CREATE_NEW_PROCESS_GROUP}
+    return {"start_new_session": True}
+
+
+def _terminate_process(proc: subprocess.Popen[str], sig: int) -> None:
+    killpg = getattr(os, "killpg", None)
+    getpgid = getattr(os, "getpgid", None)
+    if killpg is not None and getpgid is not None:
+        killpg(getpgid(proc.pid), sig)
+    elif sig == signal.SIGTERM:
+        proc.terminate()
+    else:
+        proc.kill()
 
 
 def _encode_image(image: Any) -> str:
@@ -107,14 +126,14 @@ class RemotePlanner:
             argv += ["--quantization", self._quantization]
         logger.info("RemotePlanner: launching specialist server: %s", " ".join(argv))
         # stderr inherited -> model-loading logs stay visible in the terminal.
-        # start_new_session -> own process group for clean teardown.
+        # POSIX gets its own process group; Windows falls back to direct cleanup.
         self._proc = subprocess.Popen(
             argv,
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             text=True,
             bufsize=1,
-            start_new_session=True,
+            **_popen_creation_kwargs(),
         )
         assert self._proc.stdout is not None
         self._reader = threading.Thread(
@@ -212,8 +231,8 @@ class RemotePlanner:
                 proc.stdin.close()
         with contextlib.suppress(ProcessLookupError, PermissionError, OSError):
             if proc.poll() is None:
-                os.killpg(os.getpgid(proc.pid), signal.SIGTERM)
+                _terminate_process(proc, signal.SIGTERM)
                 try:
                     proc.wait(timeout=10)
                 except subprocess.TimeoutExpired:
-                    os.killpg(os.getpgid(proc.pid), signal.SIGKILL)
+                    _terminate_process(proc, _SIGKILL)
--- a/tests/unit/test_remote_planner.py
+++ b/tests/unit/test_remote_planner.py
@@ -205,6 +205,53 @@ def test_remote_planner_roundtrip(tmp_path: Path) -> None:
         planner.close()  # idempotent
 
 
+def test_remote_planner_close_falls_back_without_posix_process_groups(monkeypatch) -> None:
+    import odyssey.runners.agents.remote_planner as remote_planner
+
+    monkeypatch.delattr(remote_planner.os, "killpg", raising=False)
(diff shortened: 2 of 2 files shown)


SoyGema (changes requested):
Thanks for taking the time on this — Windows portability is exactly the kind of contribution that's easy to overlook, and the report in #52 was really well written and reproducible. 🙏

Up front: I'm reviewing from a Linux/maintainer seat and can't fully exercise the Windows path myself, so please push back where you know the platform better than I do — you're closer to the actual failure modes here.

**What's clearly good and I'd happily take as-is:** the subprocess lifecycle work. Splitting the process-creation kwargs per platform and routing termination through `_terminate_process` (POSIX process groups via `killpg`, direct `terminate()/kill()` on Windows) is the right shape, and un-skipping the cancellation test rather than leaving it marking a known gap is the correct instinct. That half is solid.

**One CI blocker** (this is why the checks are red — it's `mypy`, not the tests): mypy doesn't treat `os.name == "nt"` as a platform guard, so on the Linux runner it still sees the Windows-only `subprocess.CREATE_NEW_PROCESS_GROUP` access and flags it. Switching those two guards (`_popen_creation_kwargs` in `remote_planner.py` and `_process_creation_kwargs` in `subprocess.py`) to `sys.platform == "win32"` lets mypy narrow the branch and the errors go away. I confirmed locally that runtime behavior on POSIX is unchanged, and that the one failing unit test is a pre-existing import-order flake unrelated to this PR.

**Where I'd love your thoughts before merging:** the paths half feels heavier than the problem. Introducing `Pathish` and widening `Path` → `Pathish` across ~6 signatures spreads a new type surface through the whole eval pipeline to satisfy some tests that assert forward slashes. And `path_preserving_style()` can hand back a `PurePosixPath` on Windows — which has no I/O (`.exists()`, `.open()`), so any downstream consumer that opens the checkpoint as a file would break silently. Could we get the same Windows fix with something narrower — e.g. keeping `Path` and normalizing at the argv-construction boundary where the separators actually matter — rather than propagating the alias? Curious whether you hit a concrete case that *needs* the pure-path return, or whether that was defensive.

One more minor note: the branch is a little behind `develop`, so a rebase would pick up the latest CI + tests.

Happy to pair on the mypy tweak if useful. Really appreciate the work here.

paranoa233:
Obsolete after force-push to `53819ef`: this comment described an earlier, broader version of the PR. The current PR body and the latest comment supersede it.

paranoa233:
Obsolete after force-push to `53819ef`: this referred to the previous `fd49fa1` CI run. The current branch now contains the narrower RemotePlanner-only fix described below.

paranoa233:
Force-pushed a narrower version after re-reading the issue guidance.

This now focuses only on the RemotePlanner cleanup failure: POSIX still uses `killpg/getpgid`, while platforms without those APIs fall back to direct `terminate()` / `kill()`. I removed the broader path-separator work from this PR because native Windows path support was called out as out of scope.

Verified locally with:
- `python -m pytest tests/unit/test_remote_planner.py -q` -> 12 passed
- `python -m ruff check src/odyssey/runners/agents/remote_planner.py tests/unit/test_remote_planner.py`
- `python -m mypy --platform linux src/odyssey`

SoyGema (approved):
Force-push nailed it — this addresses everything from the earlier review. 🙌

- **mypy blocker**: switching the guards to `sys.platform == "win32"` lets the Linux runner narrow the branch, and CI is now green across 3.10/3.11/3.12. ✅
- **Scope**: dropping the `Pathish`/path-separator half was the right call given the #52 discussion — this PR is now tightly focused on the actual RemotePlanner `close()` crash, which is exactly what I was hoping for.
- **Cleanup path**: routing teardown through `_terminate_process` (POSIX process groups via `killpg/getpgid`, direct `terminate()/kill()` where those APIs are absent) is clean, and the new regression test locks the fallback in.

One tiny non-blocking note for the record: on Windows the `CREATE_NEW_PROCESS_GROUP` child is torn down via `terminate()/kill()` on the process itself rather than the group — totally fine given native Windows support is out of scope right now, just flagging in case we revisit process-tree teardown later.

Really appreciate how responsive you were to the feedback and the clear verification notes. Approving — thanks for the solid contribution. 🚀

PR #56, opened 2026-07-08
test: isolate GR00T import guard

### LINKED ISSUE #48: Test-isolation flake: torch leaks into sys.modules, breaking test_module_imports_without_sim_deps in full-suite runs
## Summary

`tests/unit/test_gr00t_isaac_eval.py::test_module_imports_without_sim_deps` fails **only in a full-suite run** when `torch` is installed in the environment. It **passes in isolation**.

Found while validating the release PR #47 (`develop → main`). It is **pre-existing on `develop`**, not introduced by that PR.

## Symptom

```
tests/unit/test_gr00t_isaac_eval.py:51: AssertionError
E   assert mod not in sys.modules, f"{mod} leaked into module-load imports"
E   AssertionError: torch leaked into module-load imports
```

```
1 failed, 332 passed, 4 skipped
```

## Root cause

The test guards that importing the GR00T Isaac eval recipe does **not** pull in heavy sim deps (torch, isaaclab, gr00t). It asserts those modules are absent from `sys.modules`.

When the **full suite** runs and `torch` is installed locally, an **earlier test imports `torch`**, leaving it in `sys.modules`. By the time this guard test runs, `torch` is already present → the assertion fires. The recipe module itself is *not* importing torch; the pollution comes from another test in the same process.

## Why CI is green

CI installs only the `[dev]` extra (no `torch`), so nothing leaks into `sys.modules` and the guard passes. The flake only surfaces in environments where a heavy extra (e.g. `[openvla]`/`[robosuite]`) put `torch` on the path.

## Reproduce

```bash
pip install -e ".[dev]" && pip install torch    # torch present in the env
pytest -q                                        # full suite → fails
pytest tests/unit/test_gr00t_isaac_eval.py -q    # isolation → 15 passed
```

## Suggested fix

Make the guard robust to a polluted interpreter — options:
- Snapshot `sys.modules` and assert the **recipe import didn't add** the heavy modules, rather than asserting global absence; or
- Run the import check in a **subprocess** (clean interpreter) so prior tests can't pollute it; or
- `monkeypatch`/save-restore `sys.modules` around the import under test.

## Refs
- Discovered in #47 (release `develop → main`, v0.1.0-alpha)
- Affects `develop`

paranoa233:
## Summary
- run the GR00T Isaac import guard in a clean Python subprocess
- assert the recipe import itself does not load heavy sim deps, independent of parent pytest process pollution

## Tests
- `python -c "import sys, types; sys.modules['torch'] = types.ModuleType('torch'); import pytest; raise SystemExit(pytest.main(['tests/unit/test_gr00t_isaac_eval.py::test_module_imports_without_sim_deps', '-q']))"`
- `python -m pytest tests/unit/test_gr00t_isaac_eval.py tests/unit/test_imports.py -q`
- `python -m ruff check tests/unit/test_gr00t_isaac_eval.py`

Closes #48

### DIFF
--- a/tests/unit/test_gr00t_isaac_eval.py
+++ b/tests/unit/test_gr00t_isaac_eval.py
@@ -15,7 +15,9 @@ itself asserted below.
 """
 from __future__ import annotations
 
+import json
 import os
+import subprocess
 import sys
 from pathlib import Path
 from typing import Any
@@ -44,11 +46,28 @@ def _eval_task(**overrides: Any) -> EvaluationTask:
 # ---------------------------------------------------------------------------
 
 def test_module_imports_without_sim_deps() -> None:
-    # isaaclab / gymnasium / gr00t / torch are not installed in the core env;
-    # a clean import proves they are imported lazily in the run path, never at
-    # module load (otherwise this file would have failed to import at all).
-    for mod in ("isaaclab", "isaaclab_tasks", "gymnasium", "gr00t", "torch"):
-        assert mod not in sys.modules, f"{mod} leaked into module-load imports"
+    # Check import side effects in a clean interpreter so earlier tests cannot
+    # pollute sys.modules (for example by importing torch).
+    evals_dir = Path(__file__).resolve().parents[2] / "src" / "odyssey" / "runners" / "evals"
+    heavy_deps = ("isaaclab", "isaaclab_tasks", "gymnasium", "gr00t", "torch")
+    script = (
+        "import importlib, json, sys\n"
+        f"sys.path.insert(0, {str(evals_dir)!r})\n"
+        "importlib.import_module('gr00t_isaac_eval')\n"
+        f"print(json.dumps([m for m in {heavy_deps!r} if m in sys.modules]))\n"
+    )
+    result = subprocess.run(
+        [sys.executable, "-c", script],
+        capture_output=True,
+        text=True,
+        timeout=30,
+    )
+    assert result.returncode == 0, (
+        "gr00t_isaac_eval failed to import in a clean interpreter:\n"
+        f"{result.stderr}"
+    )
+    leaked = json.loads(result.stdout)
+    assert leaked == [], f"gr00t_isaac_eval imported heavy deps at module load: {leaked}"
 
 
 # ---------------------------------------------------------------------------


SoyGema (approved):
Great fix — this lands exactly on one of the options from #48 (run the guard in a clean interpreter) and does it cleanly. ✅

**What it does right:**
- Spawns a fresh `sys.executable` subprocess so prior tests can't leave `torch` (or any heavy dep) in `sys.modules` — which is the precise root cause in #48: the recipe wasn't importing torch, an earlier test in the same process was.
- Faithfully reproduces the *real* import path: `gr00t_isaac_eval` is a launched-by-path standalone script (stdlib-only at module load), and inserting the `evals` dir on `sys.path` + `import_module` matches how it's actually loaded — same as the existing top-of-file import in this test.
- Solid subprocess hygiene: `timeout=30`, `capture_output`, and surfacing `stderr` in the assertion message so a real import regression gives a readable failure rather than a bare non-zero exit.

**Verified locally on the branch:**
- `pytest tests/unit/test_gr00t_isaac_eval.py -q` → 15 passed.
- Reproduced the flake: with `torch` pre-seeded into `sys.modules`, the **old** global-absence assertion fires (`torch leaked`), while **this** version passes — so the test now actually catches module-load regressions instead of parent-process pollution.

CI is green on 3.10/3.11/3.12. Approving — thanks again @paranoa233. 🚀

SoyGema:
Thanks so much for this, @paranoa233 — two solid, well-scoped contributions back to back. 🙌

This is exactly the fix #48 was asking for: running the guard in a clean interpreter so an earlier test that imports `torch` can't leak into `sys.modules` and trip the assertion. I verified locally that it passes even with `torch` pre-seeded, while the old check fails under those same conditions — so the test now catches real module-load regressions instead of parent-process pollution.

The PR body already has `Closes #48`, so **merging this will automatically close that issue**. 🎯

Labeled and approved — really appreciate the care and the clear verification notes. 🚀

SoyGema:
cc @Danny024 — could you give this a second look when you have a moment? You own the GR00T eval side, so I'd like your eyes on it before we merge.

Context: it's a test-isolation fix for the `gr00t_isaac_eval` import guard (`Closes #48`) — runs the module-load check in a clean subprocess so an earlier test importing `torch` can't leak into `sys.modules`. No changes to the recipe itself, tests only. I've reviewed + approved and CI is green, just want the GR00T owner's sign-off. 🙏

Danny024 (approved):
This is a major fix andd isolation for the the major dependencies.

PR #58, opened 2026-07-09
feat(evals): decouple the rollout-video camera from the policy camera

### LINKED ISSUE #44: Decouple the rollout-video camera from the policy camera (dedicated eval render view)
## Context

Follow-up from #38 (rollout video capture). Raised by @Danny024: the captured clip
*"does not seem to capture the full robot with a good camera angle."* Agreed in the
PR thread to track the proper fix here as a separate enhancement.

## Why this happens today

The recorded frame **is the policy's observation image** — reusing the frame the
policy already consumes is what makes capture "near-free". In
`src/odyssey/runners/evals/robosuite.py`:

```python
_capture_frame(frames, _extract_image(obs, image_key))   # ~L319 — same frame the policy sees
```

and the eval env is built with robosuite's default `agentview` camera at 256×256
(`_make_eval_env`): `camera_names="agentview"`, `camera_height/width=256`,
`image_key="agentview_image"`.

So the clip is framed **for policy input**, not to show the robot — and you can't
pick a nicer camera via `config` without **also changing what OpenVLA sees** (the
policy expects an agentview-style view), so camera choice isn't cosmetic today.

## Decision (from the #38 discussion)

**Option 3 — render a dedicated video camera, independent of the policy camera.**
Robosuite can render multiple `camera_names` at once, so the policy keeps consuming
`agentview` while the video records a separate, well-framed camera (e.g.
`frontview`/`birdview`), at the cost of one extra render stream per step.

## Proposed approach

- Add eval `config` keys for the **video camera**, independent of the policy frame:
  e.g. `video_camera` (default `frontview`), `video_height`/`video_width` (e.g. 512).
- Build the env with **both** cameras in `camera_names` (policy cam + video cam);
  capture the *video-cam* frame for the MP4, feed the policy its *own* cam.
- **Opt-in / backward compatible**: when the new keys aren't set, fall back to the
  current behavior (record the policy frame) — no regression to the near-free path.

## Acceptance criteria

- [ ] Video frames come from a dedicated camera that frames the full robot, configurable, **without** changing the policy's input.
- [ ] Default behavior unchanged when the new keys aren't set (near-free path preserved).
- [ ] Works on the headless eval path (OSMesa / EGL).
- [ ] Documented in the README "Recording rollout videos" section.

## Refs

- Feature + discussion: #38
- Reported by @Danny024

GabLinares:
## What

Adds an **opt-in dedicated video camera** for rollout MP4s on the robosuite
eval path (incl. the multi-agent planner+pilot runtime). New eval `config`
keys: `video_camera`, `video_height` / `video_width` (default 512). Set, the
env renders a second camera for the clip while the policy keeps consuming its
own frame; unset, behavior is exactly today's near-free path.

Follow-up to #38 — implements the "option 3" decision from that thread.
Closes #44.

## Why

The recorded clip **is the policy's observation image** (`agentview` at
256×256) — reused because that's what made capture near-free, but framed for
the model, not for humans, so the robot can sit partially out of frame
(reported by @Danny024 in #38). Camera choice isn't cosmetic today: picking a
nicer camera via `config` would also change what the policy sees.

## How

- New pure helper `_resolve_cameras(cfg)` in `runners/evals/robosuite.py`
  returns the env's camera kwargs plus the obs key capture should read; used
  by both `_make_eval_env` and `run()`.
- With `video_camera` set, it is appended to `camera_names`, and robosuite's
  per-camera `camera_heights`/`camera_widths` lists render the policy cam(s)
  at their configured size and the video cam at `video_height`×`video_width`
  — one env, one extra render stream per step. The MP4 reads
  `<video_camera>_image`; the policy keeps `image_key`.
- **Opt-in / backward compatible:** keys unset ⇒ identical env kwargs and
  capture key as today. Also inert when `capture_video` is off (a second
  camera would be pure render cost).
- `video_camera` colliding with a policy camera raises a clear `ValueError`
  (robosuite renders each camera at a single resolution, so the video would
  just be the policy frame — unset the key for that).
- The multi-agent runtime path now extracts **two** frames per step: the
  planner/pilot keep the policy frame, capture takes the video frame.
- Docs: README "Recording rollout videos" gains a **"Choosing the video
  camera"** subsection; the new keys ship commented-out in both example
  missions so they're discoverable without changing example behavior.

## Acceptance criteria (from #44)

- [x] Video frames come from a dedicated, configurable camera **without**
  changing the policy's input.
- [x] Default behavior unchanged when the new keys aren't set (near-free path
  preserved).
- [x] Works on the headless eval path (verified under EGL; the change only
  alters camera kwargs + which obs key is read, nothing backend-specific —
  same backend-agnostic capture as #38, which validated OSMesa).
- [x] Documented in the README "Recording rollout videos" section.

## Test plan

- [x] Unit tests (GPU-free), 8 new: `_resolve_cameras` (defaults, inert
  without `capture_video`, appending, 512 defaults, policy-camera lists,
  collision `ValueError`) + runner wiring via a dual-camera fake env serving a
  16×16 policy frame and a 32×32 video frame — the saved GIF reads back 32×32
  while the policy (and, on the multi-agent path, the planned runtime) still
  receives 16×16.
- [x] `ruff check`, `mypy --strict`, and the unit suite pass on the develop
  base.
- [x] EGL dual-camera render check (no model): `_make_eval_env` with
  `video_camera: frontview` steps a live Lift env — obs carries 256×256
  `agentview_image` **and** 512×512 `frontview_image`.
- [x] Real GPU run: `quickstart-openvla` mission end-to-end with
  `video_camera: frontview` — the MP4 records the full robot + workspace from
  frontview while the policy consumed agentview as before. (10-step LoRA
  smoke; grade F expected — the point is the capture path, as in #38.)

## Validated on

RTX 3090 (Ubuntu, headless EGL: `MUJOCO_GL=egl`), `quickstart-openvla`
mission, single-agent OpenVLA eval path; the multi-agent path is covered by
the GPU-free runtime-injection test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -423,6 +423,27 @@ for the policy, so capture only adds a list append per step plus one encode per
 episode (offloaded to a thread). Optional knobs: `video_fps` (default 24),
 `video_format` (default `mp4`).
 
+### Choosing the video camera
+
+By default the clip records the **policy's own input frame** (`agentview` at
+256×256) — a view framed for the model, not for humans, so the robot may sit
+partially out of frame. To record from a dedicated, better-framed camera
+*without changing what the policy sees*, set `video_camera`:
+
+```yaml
+    config:
+      capture_video: true
+      video_camera: frontview    # any robosuite camera, e.g. frontview, sideview
+      video_height: 512          # optional, default 512
+      video_width: 512           # optional, default 512
+```
+
+The env then renders both cameras each step — the policy keeps consuming its
+own camera and the MP4 records the other — at the cost of one extra render
+stream per step. When `video_camera` is unset, behavior is exactly the near-free
+default above. Setting it to the policy's own camera raises an error: unset it
+instead to record the policy view.
+
 > **Dependency.** MP4 encoding needs `imageio[ffmpeg]`, which ships with the
 > `[robosuite]` (and `[all]`) extra — no extra install step if you already run
 > eval. It's best-effort: if the encoder is missing it logs a warning and the
--- a/examples/multiagent-openvla-gemma/mission.yaml
+++ b/examples/multiagent-openvla-gemma/mission.yaml
@@ -100,6 +100,9 @@ tasks:
     config:
       unnorm_key: bridge_orig
       capture_video: true   # one MP4 per episode (README: "Recording rollout videos")
+      # video_camera: frontview   # record from a dedicated camera instead of the
+      # video_height: 512         # policy's 256x256 agentview frame; unset =
+      # video_width: 512          # record the policy frame (near-free default)
 
 leaderboard:
   publish: false
--- a/examples/quickstart-openvla/mission.yaml
+++ b/examples/quickstart-openvla/mission.yaml
@@ -59,6 +59,9 @@ tasks:
     config:
       unnorm_key: bridge_orig
       capture_video: true   # one MP4 per episode (README: "Recording rollout videos")
+      # video_camera: frontview   # record from a dedicated camera instead of the
+      # video_height: 512         # policy's 256x256 agentview frame; unset =
+      # video_width: 512          # record the policy frame (near-free default)
 
 leaderboard:
   publish: false   # Backend not live yet in v0.1.0-alpha.
--- a/src/odyssey/runners/evals/robosuite.py
+++ b/src/odyssey/runners/evals/robosuite.py
@@ -115,6 +115,53 @@ def _default_env_factory(benchmark_name: str, robosuite_robot: str | None) -> An
     return robosuite.make(**kwargs)
 
 
+def _resolve_cameras(cfg: dict[str, Any]) -> tuple[dict[str, Any], str | None]:
+    """Compute the env's camera kwargs and the video-capture obs key.
+
+    Opt-in ``video_camera`` (with ``video_height``/``video_width``, default 512)
+    adds a second render camera dedicated to the rollout video, decoupled from
+    the policy's camera. Unset — or with ``capture_video`` off, where it would
+    be pure render cost — the kwargs and capture key (None) match the near-free
+    path: the video reuses the policy frame.
+    """
+    camera_names = cfg.get("camera_names", "agentview")
+    camera_height = cfg.get("camera_height", 256)
+    camera_width = cfg.get("camera_width", 256)
+    video_camera = cfg.get("video_camera")
+    if video_camera is None or not cfg.get("capture_video", False):
+        return (
+            {
+                "camera_names": camera_names,
+                "camera_heights": camera_height,
+                "camera_widths": camera_width,
+            },
+            None,
+        )
+    video_camera = str(video_camera)
+    policy_cameras = (
+        [camera_names] if isinstance(camera_names, str) else list(camera_names)
+    )
+    if video_camera in policy_cameras:
+        raise ValueError(
+            f"video_camera {video_camera!r} is already a policy camera — robosuite "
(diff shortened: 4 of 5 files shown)


GabLinares:
Example of the secondary camera in `frontview` mode at 1920x1080 resolution.

https://github.com/user-attachments/assets/a7d2ca00-7c87-42ec-aa50-67c0e7c79d76

SoyGema:
Thanks for the solid work here, @GabLinares — this is a clean, well-scoped PR and a pleasure to review. 🙏

## Overview

Adds an opt-in dedicated video camera for robosuite rollout MP4s, decoupled from the policy's input frame, via a new pure helper `_resolve_cameras(cfg)` used by both `_make_eval_env` and `run()`. Unset keys (or `capture_video` off) keep behavior byte-identical to the current near-free path.

## Code quality — strong

- **Clean design.** A single pure `_resolve_cameras` as the source of truth for both env kwargs and the capture key is the right call — consistent across call sites and trivially unit-testable (which the PR exploits well).
- **Backward compatibility is genuinely preserved.** The early return yields the exact same dict shape and `None` capture key as before; `test_resolve_cameras_defaults_unchanged_without_video_camera` locks it in with a full-dict assertion.
- **Good micro-optimization** in the runtime path: `image if capture_key == image_key else _extract_image(obs, capture_key)` avoids a redundant extraction when capture reuses the policy frame.
- **Fails fast with a clear message.** The collision `ValueError` explains *why* (single resolution per camera) and *what to do*, and it's raised before the rollout loop.
- **Conventions followed:** 3.10+ `isinstance(value, list | tuple)`, typed throughout, mypy-strict-clean.

## Test coverage — excellent

- 8 new tests covering `_resolve_cameras` (defaults, inert-without-capture, append, 512 defaults, policy-camera list expansion, collision) plus runner-level wiring via a dual-camera fake env.
- The multi-agent test (`test_video_camera_multiagent_keeps_policy_frame`) is the important one — it verifies the runtime sees only `(16,16,3)` frames while the video records 32×32, directly guarding the "two frames per step" invariant that is the trickiest part of the change. Nicely done.

## Minor observations (non-blocking)

- **Duplicate call.** `_resolve_cameras(cfg)` runs once in `_make_eval_env` and again in `run()`. Pure and cheap, so harmless — but a one-line comment noting the intentional double-call would help future readers.
- **`camera_names` vs `image_key` consistency.** Collision detection compares `video_camera` against `camera_names` (default `agentview`), while the policy reads `image_key` (default `agentview_image`). If a user sets a custom `image_key` without a matching `camera_names`, the guard checks the wrong axis. Unusual config + pre-existing coupling, but a README sentence tying `image_key` ⇄ `camera_names` would preempt confusion.
- **Uneven per-camera lists.** `_per_camera` expands a scalar to match `len(policy_cameras)` but passes a list through as-is; a wrong-length list reaches robosuite unchecked. Edge case, fine to defer.

## Risk — low

Purely additive and opt-in; the default path is provably unchanged (asserted at both helper and runner level). GPU behavior validated manually (RTX 3090, EGL, real quickstart run), CI path fully faked, docs + both example missions ship the keys commented-out.

**Verdict: Approve.** The observations above are polish, not blockers. Base is correctly `develop`. 👍

SoyGema (approved):
Approving — clean, well-scoped, backward-compatible, and thoroughly tested (esp. the multi-agent two-frames-per-step invariant). Great work, @GabLinares! 🙏 Non-blocking polish notes in the review comment above.

PR #62, opened 2026-07-13
perf(openvla): FlashAttention/SDPA + inference_mode for lower per-step latency

SoyGema:
## What

Two low-risk, **in-process** latency optimizations to the OpenVLA Pilot forward pass, which is the dominant cost in eval (it's called once per step → 50+ times per episode). No migration to a serving stack (vLLM/SGLang) and no architectural changes.

- **Faster attention at load time**: `_load_vla_model` (a helper shared by `make_openvla_policy` and `VLARuntime`) tries `flash_attention_2` if `flash_attn` is installed, otherwise `sdpa`, instead of the transformers `eager` default. **Safe fallback**: if OpenVLA's remote code (pinned transformers) rejects `attn_implementation`, it retries without the kwarg → unchanged behavior on unsupported stacks.
- **`torch.inference_mode()`** around both `predict_action` call sites to drop autograd tracking during rollouts.

No changes to: bf16, greedy decode (`do_sample=False`), single weight load. The dependency probe (`AutoModelForVision2Seq`) is kept in the import guards so that a missing `openvla` extra still raises the friendly `NotImplementedError`.

## Why

In the eval loop the Specialist is invoked once per episode but the Pilot 50+ times, so every ms shaved off the Pilot forward pass is multiplied. These are the lowest-risk levers under the "cheapen the forward pass" goal — the structural ones (episode batching, TensorRT) are deferred and depend on the OpenVLA-7B vs GR00T-3B decision.

## Scope / isolation

GR00T is a **separate** runner (`gr00t.py` / `gr00t_isaac_eval.py`) and does **not** go through this path. It only shares two utility helpers (`_flatten_config`, `_resolve_and_fetch_hf_model`) which are untouched.

## Verification

- `ruff check` clean
- `mypy` clean (strict)
- 54 GR00T + OpenVLA tests pass (including `test_gr00t*` and `test_openvla_policy`)

⚠️ **Pending GPU validation**: the real it/s impact (and confirming that `flash_attention_2` is accepted by OpenVLA's remote code with the pinned transformers version) must be measured on the L4 VM with a before/after benchmark. The fallback guarantees that, if it's not accepted, nothing breaks — it just doesn't speed things up.

### DIFF
--- a/src/odyssey/runners/models/openvla.py
+++ b/src/odyssey/runners/models/openvla.py
@@ -432,6 +432,56 @@ def _is_lora_checkpoint(checkpoint_path: Path) -> bool:
     return (checkpoint_path / "adapter_config.json").is_file()
 
 
+def _preferred_attn_implementations() -> list[str]:
+    """Attention backends to try at load time, fastest first.
+
+    FlashAttention-2 when the ``flash_attn`` package is importable, then
+    PyTorch SDPA (no extra dependency, still faster than the ``eager`` default).
+    ``_load_vla_model`` falls back to the model default if none are accepted by
+    the pinned OpenVLA remote code.
+    """
+    import importlib.util
+
+    impls: list[str] = []
+    if importlib.util.find_spec("flash_attn") is not None:
+        impls.append("flash_attention_2")
+    impls.append("sdpa")
+    return impls
+
+
+def _load_vla_model(model_ref: str) -> Any:
+    """Load an OpenVLA vision-to-seq model with the fastest available attention.
+
+    Same bf16 + ``low_cpu_mem_usage`` + ``trust_remote_code`` as before, but
+    tries FlashAttention-2 / SDPA first. If the custom OpenVLA modeling code
+    rejects ``attn_implementation`` (older pinned ``transformers``), it retries
+    without the kwarg so behaviour is unchanged on unsupported stacks.
+    """
+    import torch
+    from transformers import AutoModelForVision2Seq
+
+    common = {
+        "torch_dtype": torch.bfloat16,
+        "trust_remote_code": True,
+        "low_cpu_mem_usage": True,
+    }
+    for attn in _preferred_attn_implementations():
+        try:
+            model = AutoModelForVision2Seq.from_pretrained(
+                model_ref, attn_implementation=attn, **common
+            )
+            logger.info("Loaded %s with attn_implementation=%r", model_ref, attn)
+            return model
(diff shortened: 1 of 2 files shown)


SoyGema:
## Closing — marginal measured impact, not worth merging

Closing this PR. The two in-process levers here (FlashAttention/SDPA at load
time + `torch.inference_mode()` around `predict_action`) do **not** move
per-step latency in any meaningful way, and that is expected by design rather
than a tuning problem:

- OpenVLA rollout inference is an **autoregressive, discrete-token decode** —
  memory-bandwidth-bound, not attention-compute-bound. FlashAttention/SDPA pay
  off on long-sequence prefill; here the input is a single image + a short
  instruction, so attention is not the bottleneck. `inference_mode()` only
  removes a little autograd bookkeeping.
- The dominant eval cost is **how many forward passes we run**, not how long a
  single one takes. OpenVLA (as coded) calls the Pilot **once per env step**
  (50+×/episode), so shaving a few % off one call is negligible.

### Why the structural fix (parallel eval environments) is blocked on standard cloud GPUs

The real lever is running multiple eval environments and either batching their
observations through one model or running N model copies. On the standard
single-GPU cloud machines we target this is not currently viable:

- **VRAM ceiling.** A typical cloud eval box is a single 24 GB GPU (L4 /
  RTX-4090 class). OpenVLA-7B in bf16 is ~14 GB, and the multi-agent path keeps
  a **Specialist VLM co-resident** in the same VRAM (the two weights cannot be
  batched together — the budget is their sum). That already saturates 24 GB, so
  a **second Pilot copy does not fit** → N-copy episode parallelism is off the
  table on a single standard GPU.
- **Batched inference isn't free either.** Feeding N environments through a
  single forward pass (the throughput-efficient option) requires reworking
  OpenVLA's remote-code `predict_action`, which is single-image and pinned to
  `transformers==4.40.1`. Non-trivial, and still fights the same VRAM ceiling
  for the KV-cache/activation growth.

Multi-GPU or a larger card (≥48 GB) would relax this, but that's outside the
standard single-GPU cloud target and changes the hardware/cost story.

### Where the leverage actually is

Reducing **calls per episode**, not per-call latency — i.e. action chunking
(GR00T's action horizon = 40 → ~an order of magnitude fewer Pilot calls, plus
official TensorRT) or OpenVLA-OFT. Which of those we pursue depends on the open
Pilot-model decision (OpenVLA-7B vs GR00T-3B), so this micro-optimization is not
worth carrying.

Branch kept for reference; not merging.

Issue #61, opened 2026-07-13
LIBERO eval: finetuned OpenVLA checkpoint approaches the object but fails to grasp (0/2 episodes)


SoyGema:
Tracking a validation finding from **PR #46** (`feat/libero-eval`, LIBERO eval integration), observed running the single-agent example on a GCP L4.

https://github.com/user-attachments/assets/96304c58-511e-440d-bf2e-b51b3791ee34

## Status: eval pipeline validated end-to-end ✅
The hard part works. On `examples/franka-libero/mission.yaml` (suite `libero_object`, `task_id 0`):
- Dependency install resolved (LIBERO co-installs with the OpenVLA stack in a dedicated `env_pilot_libero`; `transformers` stays at 4.40.x, `robosuite` 1.4.0).
- `LiberoRunner` runs end-to-end: model loads, env builds, episodes run and are scored, videos are captured.

## Symptom
The published, task-specific checkpoint `openvla/openvla-7b-finetuned-libero-object` scores **0/2** episodes (`return=0.000`, `FAIL`). In the rollout videos, **the arm moves toward the object but fails to grasp it cleanly** — in both episodes. (Videos to be attached below.)

## Hypothesis
A checkpoint fine-tuned **on this exact suite** is reported at ~70–90% success, so 0/2 is suspicious. However, a code review shows the integration is **faithful to OpenVLA's reference `run_libero_eval.py` on every axis we'd expect to cause this**:

| Axis | `libero.py` | OpenVLA reference | Match |
|---|---|---|---|
| `predict_action` gripper range `[0,1]` | raw output | raw output | ✅ |
| gripper: normalize → binarize → invert | `_libero_action` | `normalize_gripper_action` + `invert_gripper_action` | ✅ |
| image 180° flip | `[::-1, ::-1]` | same | ✅ |
| camera resolution | 256×256 | 256×256 | ✅ |
| settle before acting | 10 no-ops, gripper open | `num_steps_wait` | ✅ |
| `unnorm_key` | `libero_object` | per-suite | ✅ |

So the two leading explanations are:

1. **Statistical variance** — 2 episodes is too small a sample; a ~80% checkpoint failing the grasp (the hard sub-task) twice in a row is plausible. **Action:** re-run with `num_episodes: 10` for a real `success_rate`.
[...]

PR #64, opened 2026-07-15
[MA]feat(evals): closed-loop completion-gated phase transitions for multi-agent eval

SoyGema:
## Context

The multi-agent eval runtime advanced phases **open-loop**. The SPECIALIST (Gemma) decomposed the task into sub-instructions **once** per episode; `PlannedEvalRuntime` then switched sub-instructions **blindly** after a fixed step count (`FIXED_STEPS`, default 50) or a timeout, with **no perception feedback**. Failure mode: a grasp that takes longer than the window advances to "lift" with an empty gripper and nothing notices — a likely contributor to the LIBERO grasp-but-FAIL rollouts investigated in #46 / #61.

This PR adds the "**ask if the sub-task is finished before giving the next sub-order**" behavior.

## What this adds

A new **opt-in** `COMPLETION_GATED` phase strategy: every `phase_check_every` steps the runtime asks a VLM *"is this sub-instruction satisfied in the current frame? yes/no"* and advances **only on a confirmed yes**, with `phase_max_steps` as a safety cap so a phase never wedges.

Key design decision: the detector **reuses the already-loaded out-of-process Gemma SPECIALIST** (a new `check_done` capability over the same JSON-lines IPC) — no second model, **zero extra VRAM**. It is deliberately *not* the `cosmos_reason` path, which is coupled to GR00T's Cosmos-Reason2 backbone (wrong model for an OpenVLA pilot) and is a once-per-episode narrator, not a per-step detector.

**Core**
- `runners/agents/planned.py` — `PhaseStrategy.COMPLETION_GATED`; `PhaseConfig.{check_every, max_steps_per_phase}` + `from_config()`; current frame threaded into phase-advance; detector auto-discovery (via `hasattr(check_done)`); buffered phase-advance events + `drain_phase_events()` (the sync runtime can't `await` telemetry, so the async step loop drains + emits).
- `runners/agents/runtime.py` — `CompletionDetector` protocol.
- `runners/agents/planner.py` — `LLMPlanner.check_done()` with a strict yes/no prompt + conservative `_parse_yes_no()` (ambiguous → False, so a phase never advances on an unclear answer).
- `runners/agents/planner_server.py` / `remote_planner.py` — **additive** IPC: `{"check": {...}}` → `{"done": bool}`; short `check_timeout` (5s); fail-safe to `False`.
- `runners/evals/libero.py` / `robosuite.py` — `PhaseConfig.from_config(cfg)` + `phase_advance` telemetry (with `completion` / `cap` / `fixed_steps` reason) for Episode Review.

**Example**
- `examples/franka-libero/mission-multiagent.yaml` — documents and enables the opt-in config keys.

## Backward compatibility

- Default stays `FIXED_STEPS`; `COMPLETION_GATED` is opt-in via `phase_strategy` in the mission `config:`.
- `get_action(image)` public signature unchanged; **single-agent path untouched**.
- IPC is additive — legacy `{"instruction"}` / `{"shutdown"}` intact; an old server that doesn't understand `{"check"}` replies `{"error"}` → client returns `False` → gated mode degrades safely to cap-only.
- `drain_phase_events()` returns `[]` on non-gated / non-advancing steps (zero overhead).

## How to test

**Local (no GPU) — CI-covered:**
```bash
pip install -e ".[dev]"
ruff check src/ tests/
mypy
pytest tests/unit/test_agent_runtimes.py tests/unit/test_remote_planner.py tests/integration/test_multiagent_eval.py
```
New coverage: completion-gated advancement (poll cadence, cap fallback, no-detector fallback, frame/instruction passthrough), detector auto-discovery, `_parse_yes_no` table, `LLMPlanner.check_done`, the `{"check"}` IPC path + legacy back-compat, and the `drain_phase_events()` → `phase_advance` telemetry contract.

**Opt in from a mission** (config keys, all optional; omit for the old fixed-steps behavior):
```yaml
config:
  phase_strategy: completion_gated
  phase_check_every: 10   # poll the detector every N steps (keep >= 10: ~1-2s/check)
  phase_max_steps: 100    # safety cap; force-advance if never confirmed
```

**GPU VM (real, manual — deferred from #46):** run `examples/franka-libero/mission-multiagent.yaml` (with `ODYSSEY_SPECIALIST_PYTHON` set) on the L4 and confirm:
- `phase_advance` telemetry shows `completion`-reason transitions (not just `cap`),
- both models fit in VRAM (int4 planner + bf16 pilot — the detector reuses the loaded planner),
- `success_rate` on `libero_object` `task_0` is >= the fixed-steps baseline. This is the single-vs-multi comparison deferred from #46.

## Known trade-offs / follow-ups (v2)

- The `check_done` IPC round-trip is synchronous and blocks the async step loop ~1-2s per check (bounded by `phase_check_every` + the short `check_timeout`). v2: off-thread / awaitable check.
- Others deferred: mid-episode re-planning, detector debounce/voting, request-id IPC correlation, startup capability handshake.

### DIFF
--- a/examples/franka-libero/mission-multiagent.yaml
+++ b/examples/franka-libero/mission-multiagent.yaml
@@ -63,8 +63,18 @@ tasks:
       checkpoint: openvla/openvla-7b-finetuned-libero-object   # HF repo id (eval-only)
       unnorm_key: libero_object
       task_id: 0
-      steps_per_phase: 50          # PlannedEvalRuntime: advance phase every N steps
       capture_video: true
+      # Phase advancement (PlannedEvalRuntime). Default is open-loop fixed_steps:
+      #   phase_strategy: fixed_steps   # (default) advance every steps_per_phase steps
+      #   steps_per_phase: 50
+      # Opt into CLOSED-LOOP: the SPECIALIST is re-asked "is this sub-instruction
+      # done?" from the current frame every phase_check_every steps and the phase
+      # advances only when it says yes — with phase_max_steps as a safety cap. Reuses
+      # the same loaded Gemma (no extra VRAM); adds ~1-2s per check, so keep the
+      # cadence >= 10. Meant to fix "grasped but switched to the next step too early".
+      phase_strategy: completion_gated
+      phase_check_every: 10
+      phase_max_steps: 100
       # OPTIONAL alignment contract (see mission.yaml). The declared instruction is
       # cross-checked against the benchmark's task.language; the SPECIALIST plans
       # from that same authoritative instruction. Fill from the first run's log.
--- a/src/odyssey/runners/agents/planned.py
+++ b/src/odyssey/runners/agents/planned.py
@@ -26,7 +26,11 @@ from typing import Any
 import numpy as np
 from numpy.typing import NDArray
 
-from odyssey.runners.agents.runtime import PilotRuntime, PlannerRuntime
+from odyssey.runners.agents.runtime import (
+    CompletionDetector,
+    PilotRuntime,
+    PlannerRuntime,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -34,6 +38,7 @@ logger = logging.getLogger(__name__)
 class PhaseStrategy(str, Enum):
     FIXED_STEPS = "fixed_steps"
     TIMEOUT = "timeout"
+    COMPLETION_GATED = "completion_gated"
 
 
 @dataclass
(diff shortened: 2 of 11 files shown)


SoyGema:
Pending #65 due to test without latency

SoyGema:
Superseded by #76 — *multi-agent on the GR00T pilot; drop OpenVLA MA*.

The pilot-agnostic runtime + IPC introduced here (completion-gating ) is carried forward verbatim into #76 (authorship preserved), re-pointed from the OpenVLA pilot to the chunk-emitting **GR00T** pilot via a `ChunkPilotAdapter`. OpenVLA multi-agent is dropped (single-step latency makes it unviable for production MA; the pilot decision closed to GR00T after its 10/10 LIBERO validation in #65). Closing as superseded.

PR #66, opened 2026-07-16
fix(gr00t): setup.sh py3.12 for the GR00T venv + survives read-only checkout

SoyGema:
## Context

Following the GR00T quickstart on a GCP VM, `examples/quickstart-gr00t/setup.sh` failed at the GR00T policy-server venv step:

- Isaac-GR00T's `pyproject.toml` now requires `requires-python = ">=3.12,<3.13"`, but `setup.sh` created the GR00T venv with `--python 3.10` → uv warned about the interpreter mismatch and the editable install failed.
- On this VM the Isaac-GR00T checkout is a **root-owned `/srv` clone**, so `uv venv .venv` *inside* it also failed with `Permission denied`.

## What this changes (`examples/quickstart-gr00t/setup.sh`)

- **GR00T-server venv → Python 3.12** (uv fetches a managed CPython 3.12 if the host only has 3.10). The **odyssey-core** venv stays 3.10.
- **`ODYSSEY_VENV` override** (default `.venv`): a VM already running an OpenVLA env in `.venv` can build GR00T's core elsewhere (e.g. `.venv-gr00t`) without clobbering it.
- **`GR00T_VENV_DIR` override** + **auto-relocate**: when the checkout dir isn't writable (root-owned `/srv` clone), the venv is placed at `~/gr00t-venv` instead. The editable install still targets the checkout in place; `GR00T_VENV_PYTHON` is derived from the resolved dir (so `odyssey-eval-env.sh` stays correct).

## Docs

- `constraints/gr00t-server-known-good.txt` — header bumped to py3.12. The pins were resolved on 3.10 but every version ships a cp312 wheel; flagged to **re-verify live** on the eval VM and refresh from `uv pip freeze` if anything drifts.
- `examples/quickstart-gr00t/README.md` — env table row GR00T server `3.10` → `3.12`.

## Still to do (follow-up in this PR)

- Simplify the **root `README.md`** GR00T section to just point at `setup.sh` (the current manual `pip install -e /srv/Isaac-GR00T` block is broken — it's what triggered this).
- Re-verify the 3.12 pins on the VM and commit the regenerated `uv pip freeze` if they drift.

## How to test (on the VM)

```bash
cd ~/odyssey && git fetch origin && git checkout fix/gr00t-setup-python312
# core venv named so it won't touch an existing OpenVLA .venv
ODYSSEY_VENV=.venv-gr00t ISAAC_GR00T_DIR=/srv/Isaac-GR00T bash examples/quickstart-gr00t/setup.sh
```
Expect `[2/4]` to build the GR00T venv on py3.12 (relocated to `~/gr00t-venv` if `/srv` is read-only).

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,25 +89,22 @@ cube-lift environment.
 
 **Prerequisites:**
 
-> **One-command setup** of the three eval environments (Odyssey core, GR00T
-> server, Isaac Lab) with uv: `bash examples/quickstart-gr00t/setup.sh`. See
-> [`examples/quickstart-gr00t/README.md`](examples/quickstart-gr00t/README.md) for
-> why they're separate venvs by design + the Isaac Sim provisioning. The manual
-> steps below are the step-by-step equivalent.
+> **Run `bash examples/quickstart-gr00t/setup.sh`** — it builds the three eval
+> venvs (Odyssey core, GR00T server, Isaac Lab client transport) with uv. See
+> [`examples/quickstart-gr00t/README.md`](examples/quickstart-gr00t/README.md)
+> for the full walkthrough and why they're separate venvs by design.
 
-1. Install the upstream Isaac-GR00T package — it carries the training entry
-   point (`gr00t.experiment.launch_finetune`) and the demo dataset. Accept
-   NVIDIA's weight license:
-   ```bash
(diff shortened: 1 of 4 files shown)


Danny024 (approved):
Good catch on the Python version dependencies mismatch.

Issue #74, opened 2026-07-20
Integrate Physical Intelligence π0.5 + FAST as a chunk-emitting, production-ready pilot


SoyGema:
## Context

We are relegating OpenVLA as the production pilot: its latency is structural (decode is bandwidth-bound; cost is *calls/episode*, not per-call — see the PR #62 analysis) and it emits single-step actions. The direction is **chunk-emitting, production-ready generalist pilots** — GR00T-N1.7 first (PR #65), and this issue proposes **Physical Intelligence π0.5 + FAST** as the next generalist pilot.

## Why π0.5 + FAST is a great fit for Odyssey

- **Latency (the real fix).** FAST (Frequency-space Action Sequence Tokenization) lets an autoregressive VLA emit **action chunks** efficiently → far fewer pilot calls per episode. This is exactly the "fewer calls" fix PR #62 concluded is the only structural win over OpenVLA.
- **Generalization → shrinks the OOD-phrasing problem.** π0.5 is built for open-world generalization and broad language following, so it is far less sensitive to non-native instruction phrasing than the narrowly fine-tuned OpenVLA-libero — directly de-risks the multi-agent OOD confound tracked in #73.
- **Production-readiness.** π0.5 is oriented toward real-world deployment, aligning with the production-pilot goal (see the pilot-model-decision note).
- **Fits the orchestrator design.** The regime-D orchestrator's inner loop is being designed **chunk-aware** for GR00T; π0.5 reuses that same chunk-buffer/drain + chunk-aware gating path — one runtime abstraction, two generalist pilots.

## Scope (proposed)

- **`PilotRuntime` adapter for π0.5** — buffer/drain an emitted action chunk across control-loop ticks (mirrors the GR00T chunk-aware adapter; share the abstraction, don't fork it).
- **FAST de/tokenization** — integrate the FAST action tokenizer for chunk decode.
- **Action-space + normalization mapping** — π0.5's action representation → the sim's expected action (the OpenVLA `unnorm_key` analogue).
- **Eval wiring** — `evaluation_type` / model-ref plumbing for LIBERO and robosuite, reusing the GR00T-LIBERO bridge (PR #65) plumbing where possible.
- **Chunk-aware gating** — the completion-check / hand-back cadence must operate on chunks, shared with GR00T and the orchestrator (regime D).

## Open questions (need investigation before committing)

1. **Weights + license** — which π0.5 checkpoints are openly available (openpi?) and under what license? Confirm before scoping.
[...]

PR #68, opened 2026-07-16
[MA]feat(evals): delegation-driven multi-agent eval arm (grounding)

SoyGema:
## Context

Stacked on #64. That PR gives the multi-agent eval a **closed-loop completion check** (`check_done`); this PR reuses it to add the **delegation-driven** counterpart to the planner-driven `PlannedEvalRuntime`.

Framing (research landscape §4.4 coordination regimes): #64 ≈ *planner-driven + semantic transitions*. This PR sits between the closed-loop-no-plan and full-delegation regimes: a deterministic orchestrator + the SPECIALIST as an on-demand **grounding tool** (not a sequence author) + completion-gated hand-back. It isolates one variable — **is the SPECIALIST better as sequence-author (planner) or as an on-demand grounding tool (delegation)?** — for a fair comparison on identical LIBERO tasks/metrics.

> Base is `feat/multiagent-inference-integration` so the diff shows **only** the delegation arm. Re-target to `develop` once #64 merges.

## What this adds

A new **opt-in** `coordination: delegation` regime. Instead of the SPECIALIST authoring a full plan up front, an orchestrator runs a generic `pick → place` skill template and, per phase:
1. **delegates grounding** to the SPECIALIST — *"where is the object to pick up in this scene?"* (`ground`) — and splices the returned phrase into the pilot instruction (`pick up the red mug on the left`);
2. **delegates execution** to the pilot;
3. **hands control back** when the completion check confirms the sub-task is done (`check_done`), with a step cap as a safety net.

The SPECIALIST **authors no plan** — it is a perception tool (open-weight analogue of GR-ER's spatial grounding). Default stays `coordination: planning`; delegation is opt-in.

**Zero extra VRAM:** the one loaded out-of-process Gemma answers `plan`, `check_done` **and** `ground` over the same server.

**Core**
- `runners/agents/runtime.py` — `GroundingProvider` protocol.
- `runners/agents/planner.py` — `LLMPlanner.ground()` + grounding prompt + `_parse_grounding`.
- `runners/agents/remote_planner.py` / `planner_server.py` — **additive** IPC `{"ground": {...}}` → `{"target": "..."}` (mirrors `{"check"}`); short `ground_timeout`; fail-safe echoes the query; back-compat with an old server (no `target` key → echo query).
[...]

PR #65, opened 2026-07-16
feat(evals): GR00T pilot on LIBERO (single-agent eval bridge)

SoyGema:
## What

Adds an evaluation path that runs an **NVIDIA GR00T N1.7 pilot on the LIBERO benchmark**, selected via `config: {pilot: gr00t}` on an `evaluation_type: libero` task. **Single-agent only** — multi-agent chunk-aware gating is a deliberate follow-up.

GR00T emits an action **chunk** executed open-loop, so the pilot is queried ~an order of magnitude fewer times per episode than the single-step OpenVLA pilot, and the 3B model fits a standard 24 GB cloud GPU. This is the structural latency win that #62 (OpenVLA per-step micro-opts, closed as marginal) could not deliver, and it enables an apples-to-apples **GR00T vs OpenVLA on the same LIBERO suite** — closing the open Pilot-model decision.

## Result — ✅ validated end-to-end on an L4

`odyssey run examples/franka-libero/mission-gr00t.yaml` → **success_rate 1.000 (10/10) on `libero_object` task 0**, GR00T serving the checkpoint offline and executing action chunks in LIBERO/MuJoCo. Rollout mp4s captured. The GPU validation this PR was blocked on is done.

## How

Clones the proven **subprocess + policy-server** pattern from the Isaac Lab GR00T recipe rather than an in-process pilot: the heavy GR00T model runs in a separate server (`gr00t.eval.run_gr00t_server`); this process holds only the lightweight `PolicyClient` + the LIBERO env, so its dependency world stays LIBERO. No Isaac Sim — LIBERO runs on MuJoCo.

- **`_gr00t_server.py`** — server lifecycle (`serve_checkpoint` / `wait_for_server` / `build_server_command` / `connect_policy_client`). `server_python` now defaults to the setup's GR00T venv (`$GR00T_VENV_PYTHON` → `$ISAAC_GR00T_DIR/.venv/bin/python`), so the mission needs no manual wiring. **No behavior change to Isaac eval.**
- **`gr00t_libero_eval.py`** — the LIBERO eval script: env + chunked action loop + `ODYSSEY_*` protocol. `_resolve_served_path` resolves an HF repo id to its local cache snapshot (offline) and serves the per-suite subdir, so `checkpoint: nvidia/GR00T-N1.7-LIBERO` works out-of-the-box.
[...]

PR #76, opened 2026-07-21
[MA] feat(evals): multi-agent on the GR00T pilot; drop OpenVLA MA (supersedes #64/#68)

SoyGema:
## What

Consolidates **multi-agent eval onto the GR00T chunk-emitting pilot** and **removes OpenVLA multi-agent** (OpenVLA stays single-agent). **Supersedes #64 (planner + completion-gating) and #68 (delegation)** — their pilot-agnostic runtime + IPC is carried forward, re-pointed at GR00T.

> Builds on **#65** (GR00T single-agent LIBERO bridge), now merged to `develop`. Targets `develop`; the diff is only the MA work.

## Why

OpenVLA is not viable for production multi-agent: single-step, so an episode costs N pilot calls **×** the SPECIALIST calls, with no chunking. The pilot decision closed to **GR00T** (validated 10/10 on LIBERO in #65). GR00T emits action **chunks** (~10× fewer pilot calls/episode) and, as a generalist, is far less sensitive to the non-native phrasing the SPECIALIST feeds it — which also *reduces* the OOD confound in the planner-vs-delegation study. Kept pilot-agnostic so **π0.5** slots in next.

## Design

- **`ChunkPilotAdapter`** (`runners/evals/chunk_pilot_adapter.py`, new) — makes a chunk-emitting VLA look like a single-step `PilotRuntime`: buffers a chunk, drains one action per `act()`, and **re-queries when the instruction changes**. That instruction-change flush **is the chunk-aware gating** — when a runtime advances a phase it feeds a new sub-instruction, the adapter discards the stale chunk and re-queries. Pilot-agnostic: model specifics live in injected `obs_builder`/`action_mapper` callables (π0.5 = different transforms, same buffering). The full obs is pushed via `set_obs()` so the runtimes + `PilotRuntime` protocol stay env-/pilot-agnostic (rejected: widening `get_action` to a full obs dict).
- **MA runs inside the recipe** (`gr00t_libero_eval.py`) — which already owns the GR00T server + LIBERO env. `coordination: planning|delegation` builds `PlannedEvalRuntime`/`DelegatedEvalRuntime` with the adapter as pilot + `RemotePlanner` as SPECIALIST; one `env.step` per `runtime.get_action`. Single-agent path unchanged.
- **`libero.py`** — `build_gr00t_libero_argv` forwards the runner-derived SPECIALIST (`--specialist_model/quantization/python` from the loadout + `ODYSSEY_SPECIALIST_PYTHON`) when `coordination` is set; `coordination`/`phase_*` pass through as config. Dropped the OpenVLA `PlannedEvalRuntime` dispatch + `_build_planned_runtime`.
- **`robosuite.py`** — dropped its OpenVLA multi-agent path + dead helpers (single-agent only).
- **Carried from #64/#68**: `planned.py` (completion-gating), `delegated.py`, `runtime.py` (`CompletionDetector`/`GroundingProvider`), `RemotePlanner` `check`/`ground` IPC — no OpenVLA binding.

## Config surface

`examples/franka-libero/mission-gr00t-multiagent.yaml`: GR00T PILOT + Gemma SPECIALIST loadout; `pilot: gr00t` + `coordination: planning` (or `delegation`) + `phase_strategy: completion_gated`. Run with `ODYSSEY_SPECIALIST_PYTHON` set. Absent `coordination` → single-agent (back-compat with `mission-gr00t.yaml`).

## Verification

- `ruff` + `mypy(strict)` clean; **453 unit tests pass** — incl. new `ChunkPilotAdapter` tests (buffer drain / re-query / **flush-on-instruction-change** / `set_obs` threading / reset) and MA argv-forwarding tests; the obsolete OpenVLA-MA dual-camera video test was removed.
- ⚠️ **GPU smoke pending (draft).** Highest risk: NVIDIA's **sim-policy-wrapper** was built for open-loop chunk replay; MA re-queries mid-episode on phase changes — verify no internal step-counter drift (may need `sim_policy_wrapper: false` for the MA path). Then a full `libero_object` run, `planning` then `delegation`, checking `phase_advance` shows real (not cap-only) transitions.

## Scope

- OpenVLA single-agent kept; OpenVLA multi-agent removed everywhere (libero + robosuite).
- π0.5 is the next pilot on the same adapter (issue #74).


SoyGema:
## Added: orchestration arm (regime D) — LLM router

#76 now carries **all three** multi-agent coordination modes on GR00T:

| `coordination` | Who owns the sequence | SPECIALIST role |
|---|---|---|
| `planning` | SPECIALIST authors it up front (`plan`) | planner |
| `delegation` | fixed `pick → place` template | grounding tool (`ground`) |
| **`orchestration`** (new) | **an LLM ORCHESTRATOR decides it dynamically** (`route`) | router + judge |

**Regime D** = the LLM router we designed: at each phase boundary it looks at the live scene + what's already done and emits the next sub-instruction (or DONE); `check_done` gates hand-back. New `role: ORCHESTRATOR` (id `router`), served by the **same reused Gemma** (new `route` IPC verb alongside `plan`/`check`/`ground` — zero extra VRAM), driving the **same `ChunkPilotAdapter`** over GR00T. A `max_phases` cap guards runaway routing; `route` fails safe to the task (never a spurious done).

New: `OrchestratorRuntime`/`RouteDecision` (runtime.py), `OrchestratedEvalRuntime` (orchestrated.py), `examples/franka-libero/mission-gr00t-orchestration.yaml`. Tests: runtime (route/hand-back/history/done/max_phases/cap/fail-safe) + `route` IPC roundtrip + `serve()` handler. All green (469 unit tests, ruff + mypy clean).

PR #86, opened 2026-08-03
feat(providers): register ur10e embodiment (GR00T NEW_EMBODIMENT)

SoyGema:
## What

Registers **`ur10e`** in the local robot catalog (`providers/local/robots.py`) so a
`robot.embodiment: ur10e` mission spec validates end to end.

## Why

`ur10e` is the arm for the UR10e drug-sort GR00T pilot. It is driven by the
**`gr00t` runner** (`NEW_EMBODIMENT` finetune), **not** Robosuite — which is why it
was previously excluded from the catalog (the trim note warns that Robosuite-only
arms produced misleading default-Panda eval runs). This change integrates the
embodiment **unitarily and independently** of the larger drug-sort PoC branch, so
missions can name `ur10e` without pulling in the whole pilot stack.

## Changes

- `src/odyssey/providers/local/robots.py` — add `ur10e` to `KNOWN_EMBODIMENTS`, and
  rewrite the catalog comment: the allowlist is now a **superset** of the
  Robosuite-drivable arms **plus** GR00T `NEW_EMBODIMENT` finetune targets, not a
  mirror of `ROBOSUITE_ROBOT_NAMES`.
- `tests/unit/test_local_providers.py`:
  - Evolve the catalog invariant: Robosuite-translatable names are now a
    **subset** of the catalog, and `ur10e` is the sole GR00T-only addition
    (`KNOWN_EMBODIMENTS - ROBOSUITE_ROBOT_NAMES == {"ur10e"}`).
  - Add a positive resolution test (`ur10e` resolves through `LocalRobotProvider`).
  - Add `test_ur10e_resolves_then_robosuite_refuses_it`, pinning the behaviour
    change: `ur10e` now resolves cleanly and only a Robosuite eval refuses it —
    loudly (`ValueError`), never a silent default-Panda substitution.
- `docs/concepts.md` — `embodiment:` row updated 8 → **9 names**; note that `ur10e`
  resolves and validates but a Robosuite eval task against it **raises** rather than
  passing through to `robosuite.make(robots=...)`.
- `src/odyssey/runners/evals/robosuite.py` — amend the `ROBOSUITE_ROBOT_NAMES`
  comment: it is the Robosuite-drivable **subset** of `KNOWN_EMBODIMENTS`, the
  invariant this PR deliberately breaks.

## Test

`pytest tests/unit/test_local_providers.py` → **13 passed** (incl. the two new
ur10e tests and the updated invariant). Full suite **499 passed, 5 skipped**;
`ruff check` and `mypy` strict clean on 3.10–3.12.

## Notes

- No Robosuite translation is added for `ur10e` (Robosuite doesn't ship it); the
  guard that every Robosuite name stays in the catalog is preserved as a subset.
- Scoped intentionally small: catalog + tests + docs only. The finetune mission,
  modality config, and eval live on the `feat/ur10e-drugsort-finetune` branch.
- The gate this PR actually changes is `create_mission` → `LocalRobotProvider.resolve`
  (not `odyssey validate`, which only loads the spec and never consults the catalog —
  it already printed `OK` for `embodiment: ur10e` on `develop`).

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -109,7 +109,7 @@ and that `agents` contains exactly one agent.
 
 | Form | Example | What it does today |
 |---|---|---|
-| `embodiment:` | `franka_panda`, `ur5e`, `sawyer` | Names a built-in catalog embodiment. 8 names accepted: `franka_panda`, `panda`, `sawyer`, `iiwa`, `jaco`, `kinova_gen3`, `ur5e`, `baxter` — the arms Robosuite's built-in robot models cover. Resolved at mission-creation by `LocalRobotProvider`; passed through to `robosuite.make(robots=...)` at evaluation. |
+| `embodiment:` | `franka_panda`, `ur5e`, `ur10e` | Names a built-in catalog embodiment. 9 names accepted: `franka_panda`, `panda`, `sawyer`, `iiwa`, `jaco`, `kinova_gen3`, `ur5e`, `baxter` — the arms Robosuite's built-in robot models cover, passed through to `robosuite.make(robots=...)` at evaluation — plus `ur10e`, driven by the **GR00T runner** (`NEW_EMBODIMENT` finetune), which Robosuite doesn't ship: it resolves and validates, but a Robosuite eval task against it raises rather than passing through. All resolved at mission-creation by `LocalRobotProvider`. |
 | `urdf:` | `./arms/my_arm.urdf` | Names a local URDF/xacro path. Existence-checked at mission-creation. No robot pass-through to Robosuite — falls back to the env's default robot. |
 | `id:` | `rob_01HQR...` | Reserved for a robot registered in your Lovell account. When the Lovell provider ships, this will fetch the loadout from the account rather than requiring an inline `agents:` block. Requires `odyssey login`, not yet shipped. |
 
--- a/src/odyssey/providers/local/robots.py
+++ b/src/odyssey/providers/local/robots.py
@@ -28,14 +28,19 @@ from odyssey.providers.base import ResolvedRobot, RobotProvider
 from odyssey.spec.mission import RobotSpec
 
 # Scoped to embodiments at least one shipped runner can actually drive
-# end-to-end. Today that means the arms Robosuite's built-in robot
-# models cover (``ROBOSUITE_ROBOT_NAMES`` in runners/evals/robosuite.py is the
-# matching translation table). Quadrupeds (unitree_go2/h1), mobile bases
-# (tiago, stretch3), and arms Robosuite doesn't ship (ur10e) were
-# removed in this trim — they accepted with a green spec, then produced
-# identical default-Panda eval runs that misled users about what was
-# being simulated. Re-add a name here only when a runner exists that
-# honors it; for unsupported robots, ``urdf:`` still works.
+# end-to-end. Two runners qualify a name today:
+#   * Robosuite — the arms its built-in robot models cover
+#     (``ROBOSUITE_ROBOT_NAMES`` in runners/evals/robosuite.py is the matching
+#     translation table).
+#   * the GR00T runner — arms it can finetune as a NEW_EMBODIMENT (e.g.
+#     ``ur10e``), which Robosuite doesn't ship and so stays outside
+#     ``ROBOSUITE_ROBOT_NAMES``.
+# The catalog is therefore a superset of the Robosuite-drivable names, not a
+# mirror of them. Quadrupeds (unitree_go2/h1) and mobile bases (tiago,
+# stretch3) remain excluded — with no runner that honors them they accepted
+# with a green spec, then produced identical default-Panda eval runs that
+# misled users about what was being simulated. Re-add a name here only when a
+# runner exists that drives it; for unsupported robots, ``urdf:`` still works.
 KNOWN_EMBODIMENTS: frozenset[str] = frozenset(
     {
         "franka_panda",   # alias of "panda" — common in OpenVLA / LeRobot specs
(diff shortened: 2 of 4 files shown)


Danny024 (changes requested):
Hi Gema — nicely scoped PR, and thank you for unbundling this from the PoC branch.

I pulled the branch and ran it locally: `ruff check` clean, `mypy` strict clean, full suite **499 passed**, and your 12 targeted tests pass exactly as described. CI is green on 3.10–3.12.

I particularly liked that you *evolved* the catalog invariant instead of deleting it — keeping `⊆` plus an explicit exception set preserves both failure directions, which the easy version would have lost.

## Blocking (documentation only — no code change needed)

**1. `src/odyssey/providers/local/robots.py:34`** — the comment immediately above your addition still reads *"arms Robosuite doesn't ship (**ur10e**) were removed in this trim"*. The file now names `ur10e` as removed three lines above it being present. Line 31 (*"the arms Robosuite's built-in robot models cover"*) is stale for the same reason. Could you update the paragraph to the new rule: Robosuite-drivable arms **plus** GR00T finetune targets?

**2. `docs/concepts.md:112`** — says *"8 names accepted"* and lists them; it's 9 now, and the *"the arms Robosuite's built-in robot models cover"* qualifier no longer holds. The same row also promises embodiments are *"passed through to `robosuite.make(robots=...)` at evaluation"*, which for `ur10e` raises.

## Non-blocking, worth a look

**3.** Behaviour change I'd like pinned with a test: on `develop`, `embodiment: ur10e` + a Robosuite eval was rejected at `create_mission`; now it's accepted and fails at task-run time. It still fails *loudly* — `_resolve_robosuite_robot` raises, no silent Panda substitution — but `test_unsupported_embodiment_fails_with_listing` uses `model_construct` to bypass the validator precisely because no real catalog name could reach that path. `ur10e` is the first one that can. This passes today:

```python
async def test_ur10e_resolves_then_robosuite_refuses_it() -> None:
    spec = RobotSpec(embodiment="ur10e", agents=_agents())   # no model_construct needed now
    assert (await LocalRobotProvider().resolve(spec)).embodiment == "ur10e"
    with pytest.raises(ValueError, match="Robosuite has no built-in robot"):
        _resolve_robosuite_robot(spec)
```

**4.** `src/odyssey/runners/evals/robosuite.py:76` claims the map *"mirrors the trimmed `LocalRobotProvider.KNOWN_EMBODIMENTS` so the same names that pass spec validation also run end-to-end"* — that's the invariant this PR deliberately breaks. One-line amendment.

**5.** Minor note on the description: `odyssey validate` already printed `OK` for `embodiment: ur10e` on `develop` — it only loads the spec and never consults the catalog. The gate you've actually changed is `create_mission` → `LocalRobotProvider.resolve`. Real change, just a different gate than the body names.

Happy to approve as soon as 1 and 2 land.

SoyGema:
@Danny024 thanks for the thorough review — pushed `0a0837e` addressing all points.

**Blocking (docs):**
1. `robots.py` — rewrote the `KNOWN_EMBODIMENTS` paragraph. It no longer names `ur10e` as "removed in this trim"; the new rule is spelled out: the catalog is a **superset** of the Robosuite-drivable arms **plus** GR00T `NEW_EMBODIMENT` finetune targets. The stale "arms Robosuite's built-in robot models cover" line is gone.
2. `docs/concepts.md` — 8 → **9 names**; the row now says `ur10e` resolves/validates but a Robosuite eval against it **raises** rather than passing through to `robosuite.make(robots=...)`.

**Non-blocking — done as well:**
3. Added `test_ur10e_resolves_then_robosuite_refuses_it` (your snippet, verbatim behaviour) — pins that `ur10e` resolves and only Robosuite refuses it, loudly, no silent Panda substitution.
4. `robosuite.py:76` comment amended — `ROBOSUITE_ROBOT_NAMES` is now described as the Robosuite-drivable **subset** of the catalog, not a mirror.
5. Good catch on the gate — updated the PR description: the real gate is `create_mission` → `LocalRobotProvider.resolve`, not `odyssey validate`.

Full suite **499 passed, 5 skipped**; `ruff` + `mypy` strict clean. Ready for another look 🙏

Danny024 (approved):
Thanks Gema — all four items addressed, and the new commit is DCO-signed. I re-pulled the branch and verified rather than eyeballed:

| Check | Result |
|---|---|
| `ruff check src tests` | ✅ clean |
| `mypy src/odyssey` (strict) | ✅ 78 files, no issues |
| `pytest tests/unit/test_local_providers.py` | ✅ **13 passed** (was 12) |
| `pytest -q` full suite | ✅ **500 passed** (was 499) |
| CI 3.10 / 3.11 / 3.12 | ✅ all green |

**The new test earns its place.** I mutation-tested it — patched `_resolve_robosuite_robot` to `return None` (silent default-Panda) instead of raising, and `test_ur10e_resolves_then_robosuite_refuses_it` failed with `DID NOT RAISE ValueError`, then passed again on restore. It genuinely pins the no-silent-substitution guarantee rather than restating the code.

**Docs verified programmatically**, not by reading:

```
docs claims      : 9 names        catalog actually : 9 names
docs enumerates  : 8 + ur10e = 9  count agrees     : YES
set agrees       : YES            robosuite subset : YES
```

The enumeration in `concepts.md` is exactly `KNOWN_EMBODIMENTS`, and the new sentence about a Robosuite eval raising is accurate.

Also appreciated that you corrected the gate in the PR body (`create_mission` → `LocalRobotProvider.resolve`, not `odyssey validate`) — that was item 5 and I'd flagged it as optional.

**Approving.**

One nit, purely optional and not worth another push on its own — `runners/evals/robosuite.py`, the second paragraph now reads:

> Anything missing here either has no Robosuite equivalent (Unitree quadrupeds, Tiago, Stretch3; **GR00T-only arms like ur10e**) or hasn't been wired yet. For those, operators can use `urdf:` and we'll fall through to Robosuite's default robot.

Adding ur10e to that list points it at the `urdf:` remedy, whereas the paragraph directly above correctly says ur10e runs under the GR00T runner. Worth untangling next time you're in the file.

PR #81, opened 2026-07-22
feat(pi05): chunk-emitting π0.5 pilot + FAST scoping (issue #74)

SoyGema:
# feat(pi05): chunk-emitting π0.5 pilot + FAST scoping (issue #74)

Integrates **Physical Intelligence π0.5 (pi05)** as the next chunk-emitting,
production-oriented generalist pilot, alongside GR00T-N1.7. Implements the
odyssey-side wiring end-to-end for a **single-agent LIBERO eval** and resolves the
open scoping questions from #74. **The GPU acceptance run has now passed on a GCP
L4 — π0.5 scores real successes on `libero_object` (see _Acceptance run_).**

Relates to #74. Reuses the chunk-aware runtime built for GR00T (PR #65) rather than
forking it.

## Design decision: server-side (option A), externally-served

π0.5 is a **flow-matching** VLA: at inference it emits **continuous** action chunks,
not FAST tokens. So the model runs out-of-process in an **openpi
`WebsocketPolicyServer`** and this repo only holds the lightweight websocket client
+ the LIBERO env. Consequences, made explicit:

- **FAST is NOT on the inference path.** FAST (de)tokenization matters only for π0.5
  *training* or for an autoregressive *π0-FAST* model. The `pi05_fast.py` codec is
  implemented + unit-tested but documented as **reserved scaffolding**, not wired.
- **Externally-served.** Unlike GR00T (which can self-serve its server in-process),
  you start the openpi server yourself and point the mission at `host:port`. A
  `_pi05_server.py` auto-serve lifecycle is intentionally deferred.

## What's in this PR

**Pilot runtime & wire glue**
- `runners/agents/chunk_pilot.py` — `ChunkPilotAdapter`: pilot-agnostic chunk
  buffer + replay + flush-on-instruction-change gating, factored out of GR00T's
  inlined loop so π0.5 reuses it (no duplicate abstraction).
- `runners/models/pi05.py` — `make_pi05_pilot()`: wraps the openpi client's
  `infer` in the adapter (lazy `openpi_client` import; injectable client for tests).
- `runners/evals/pi05_transforms.py` — openpi `observation/*` wire format + 7-DoF
  action slice, **no gripper fix-up** (contrast GR00T), reusing
  `quat_xyzw_to_axis_angle` from `gr00t_transforms`.
- `runners/evals/pi05_norm.py` — the `unnorm_key` analogue: π0.5 bakes
  `norm_stats` into the checkpoint (asset_id), so this mirrors openpi's
[...]

PR #88, opened 2026-08-03
feat(evals): CustomEvalRunner for evaluation_type: custom (#87)

SoyGema:
Closes #87.

## What

Adds **`CustomEvalRunner`**, the eval runner for `evaluation_type: custom`. It runs
an arbitrary eval script as a subprocess, decoupled from any sim.

## Why

A mission requires exactly one eval task, but only `robosuite` and `isaac_lab`
runners existed — `evaluation_type: custom` fell through to the wildcard
**`CPUMockRunner`** (a *fake* eval). There was no first-class way to run a real,
project-specific evaluation. This bites the **ur10e/ur5e drug-sort GR00T pilot**,
whose honest metrics are ground-truth evals (open-loop MAE / closed-loop lift+seat),
not a sim-to-sim gap. The only prior options were to **abuse** the `isaac_lab`
contract or to **mock** the eval — both hacks.

## Contract (the two open questions in the issue, decided + documented)

- **Serving → the script owns it.** The runner passes `--checkpoint` and never
  serves the policy. The motivating open-loop GT eval loads the checkpoint
  in-process; delegating keeps `CustomEvalRunner` free of any GR00T/openpi/Isaac/
  MuJoCo dependency. (No `config.serve_checkpoint` — a closed-loop eval starts or
  connects to a server from inside the script.)
- **Metrics → `--out-json`.** The runner allocates a path under the task output
  dir and passes `--out-json`; the script writes a JSON object there.
  `success_rate` (when present) yields a trainer letter grade + pass/fail;
  otherwise the metrics are surfaced as-is — a metric-only eval (per-joint MAE) is
  **not** forced into a fabricated `0.0 / grade-F`.

Launch: `<eval_python> <eval_script> --checkpoint <path> --out-json <path>
[--<config-key> <value> …]`. `eval_python` defaults to `sys.executable` (so the
script can run under a separate venv without Odyssey importing its deps); extra
`config.*` keys pass through verbatim, snake_case preserved — same style as
`isaac_lab`, minus the Isaac-specific `--task/--headless`.

## Changes

- `src/odyssey/runners/evals/custom.py` — new runner + `resolve_eval_script` /
  `resolve_interpreter` / `build_custom_argv` / `read_metrics` / `summarize`.
  Full contract in the module docstring.
- `src/odyssey/cli/commands/run.py` — register `CustomEvalRunner` before the CPU
  mock, so `(EVALUATION, "custom")` matches it, not the wildcard fallback.
[...]

PR #90, opened 2026-08-03
feat(pi05): π0.5 (openpi) training runner + fine-tune example

### LINKED ISSUE #74: Integrate Physical Intelligence π0.5 + FAST as a chunk-emitting, production-ready pilot

SoyGema:
Closes #74 — the last leg of the π0.5 pilot integration (pilot + eval landed in #81; this adds fine-tuning). Merging this closes the issue.

## What

Adds `Pi05Runner`, the **fine-tuning** counterpart to the existing π0.5 inference/eval path (issue #74). π0.5 now covers training + eval + inference in Odyssey.

## Why openpi looks different from GR00T/OpenVLA

GR00T and OpenVLA take a flat bag of `--flag value` overrides. **openpi is config-name-driven**: training selects a registered `TrainConfig` (bundling data pipeline + weight loader + optimizer), and CLI `--overrides` (tyro) tweak fields. So the runner:

- shells out to **two openpi entry points in order** — `scripts/compute_norm_stats.py` then `scripts/train.py` (openpi requires norm stats before training; skippable via `compute_norm_stats: false`);
- runs both with `cwd = <task output_dir>` so openpi's cwd-relative `./assets` and `./checkpoints` land under the Odyssey task dir (no guessing an upstream `--checkpoint-base-dir` flag);
- forwards mission `config:` as tyro overrides — nested dotted flags (`data.repo_id` → `--data.repo-id`), **booleans as toggle flags** (`--overwrite` / `--no-overwrite`), not `--flag True`;
- for a **local absolute** dataset ref, sets `HF_LEROBOT_HOME` to the parent dir so openpi's LeRobot loader resolves `data.repo_id` == the folder name.

Routing: registered behind OpenVLA's wildcard training slot, reached via `config: {runner: pi05}` — same mechanism as `gr00t`.

## Changes

- `src/odyssey/runners/models/pi05_train.py` — `Pi05Runner` + `build_pi05_train_argv` / `build_pi05_norm_stats_argv` / `_lerobot_env_for_dataset` / `_resolve_output_checkpoint` / `parse_pi05_train_line`. Kept **strict-clean in a new module** (not in the mypy-exempt `pi05.py` pilot file) since it's pure subprocess glue with no openpi imports.
- Registered in `runners/__init__.py` + `cli/commands/run.py`.
- `examples/quickstart-pi05-train/` — base fine-tune mission wired for the UR10e drug-sort v0 LeRobot demos (7-DoF, exterior + wrist cams) + README with the openpi `TrainConfig` recipe. Eval task is `custom` (honest closed-loop vs `meta/gt`); README explains why LIBERO is **not** an honest metric for a UR10e-fine-tuned model.
- `tests/unit/test_pi05_train.py` — 25 tests (argv builders, env overlay, checkpoint resolver, stdout parser).

## Verification

- ruff clean · mypy strict clean (79 files) · **525 passed, 4 skipped**
- `odyssey validate` + `odyssey run --use-mock-runner` pass end-to-end on the new example.

## Known gap

A first GPU run surfaced the openpi CLI-contract fixes now on this branch (the `compute_norm_stats.py --config-name` flag vs `train.py`'s positional `config_name`, `--data.repo-id`, cwd-relative `./assets`/`./checkpoints`) — those are verified against upstream source. **Still to confirm on GPU:** a full fine-tune-to-convergence run end-to-end. Same posture as the π0.5 eval path landed with.

### DIFF
--- a/examples/quickstart-pi05-train/README.md
+++ b/examples/quickstart-pi05-train/README.md
@@ -1,0 +1,102 @@
+# π0.5 fine-tune quickstart (training)
+
+Fine-tune a Physical Intelligence **π0.5** (openpi) checkpoint on a LeRobot-format
+demonstration set through the odyssey `pi05` training runner. The runner shells
+out to openpi's own entry points — `scripts/compute_norm_stats.py` then
+`scripts/train.py` — so the actual training happens in your openpi checkout.
+
+This is the **training** counterpart to `examples/quickstart-pi05/` (eval-only).
+
+> **Status:** the odyssey-side wiring (argv build, norm-stats + train subprocess
+> orchestration, checkpoint capture, stdout→progress parsing) is complete and
+> unit-tested on CPU. It has **not** been run end-to-end on a GPU box yet —
+> treat the openpi flag names below as the contract to confirm on the first smoke.
+
+## What you need
+
+- A GPU box (π0.5 fine-tune is memory-hungry — an 80 GB A100/H100 class card).
+- **openpi** installed and checked out:
+  ```bash
+  git clone https://github.com/Physical-Intelligence/openpi
+  cd openpi && pip install -e .
+  export OPENPI_REPO_PATH=$(pwd)      # runner defaults to /srv/openpi otherwise
+  ```
+- The LeRobot dataset unpacked on the box, e.g.
+  `/data/ur10e_drugsort_v0/ur10e_partial_cond_aug/`.
+
+## Why openpi is config-name-driven (and GR00T isn't)
+
+GR00T / OpenVLA take a flat bag of `--flag value` overrides. openpi selects a
+**registered `TrainConfig`** by name — it bundles the data transforms, the base
+weight loader and the optimizer. So fine-tuning a *new* dataset/embodiment means
+adding one config entry (the π0.5 analogue of GR00T's `modality_config_path`).
+
+### 1. Register a `TrainConfig` in openpi
+
+Add an entry to `src/openpi/training/config.py` (name it `pi05_ur10e_drugsort`, to
+match `config_name` in the mission). Base it on the shipped `pi05_libero` config,
+then point its data config at your LeRobot dataset and describe the state/action
+mapping:
+
+- **state/action**: 7-DoF (6 joints + gripper) — π0.5 pads to the model's action
+  dim; no observer-conditioned `grasp_target` channel (that's GR00T-specific —
+  drop it here).
+- **images**: map `observation.images.exterior` → `base_0_rgb` and
+  `observation.images.wrist` → `left_wrist_0_rgb` in the repack transform.
+- **repo_id**: the dataset folder name, `ur10e_partial_cond_aug`.
+
+### 2. Dataset resolution
+
+For a **local absolute** dataset ref, the runner sets `HF_LEROBOT_HOME` to the
+dataset's parent dir, so openpi's LeRobot loader resolves `data.repo_id` == the
+folder name. Keep the two in sync:
+
+```
+ref:            /data/ur10e_drugsort_v0/ur10e_partial_cond_aug
+data.repo_id:   ur10e_partial_cond_aug          # HF_LEROBOT_HOME=/data/ur10e_drugsort_v0
+```
+
+## Run
+
+```bash
+# CPU validation (no training):
+odyssey validate mission.yaml
+odyssey run mission.yaml --use-mock-runner
+
+# Real fine-tune on the GPU box:
+export OPENPI_REPO_PATH=/path/to/openpi
+odyssey run mission.yaml
+```
+
+The runner executes, in order:
+
+1. `python scripts/compute_norm_stats.py pi05_ur10e_drugsort`
+   (skip with `config: {compute_norm_stats: false}` if `./assets` already holds them)
+2. `python scripts/train.py pi05_ur10e_drugsort --exp-name finetune-pi05-ur10e --overwrite \
+      --data.repo-id ur10e_partial_cond_aug --num-train-steps 30000 --batch-size 32`
+
+Both run with `cwd` = the task's `output_dir`, so openpi's cwd-relative `./assets`
+and `./checkpoints` land under the odyssey run dir. The trained checkpoint is
+captured from `checkpoints/<config_name>/<exp_name>/<step>/` (highest step).
+
+## Config keys the runner interprets
+
+| key                  | meaning                                                            |
+|----------------------|-------------------------------------------------------------------|
+| `runner: pi05`       | routes the wildcard training task to `Pi05Runner`                 |
+| `config_name`        | **required** — the registered openpi `TrainConfig` (positional)   |
+| `exp_name`           | openpi `--exp-name`; defaults to the task name                     |
+| `overwrite`          | emit `--overwrite` (default `true`); clobbers the prior exp dir    |
+| `resume`             | emit `--resume` instead of `--overwrite`; continue latest ckpt    |
+| `compute_norm_stats` | run the norm-stats pre-step (default `true`)                       |
+| *anything else*      | forwarded as a tyro override (`a_b` → `--a-b`; nested `x: {y: 1}` → `--x.y 1`; bools → `--flag` / `--no-flag`) |
+
+## Can I then evaluate on LIBERO?
+
+Not honestly with *this* checkpoint. LIBERO is a **Franka Panda** sim benchmark;
+a π0.5 fine-tuned on **UR10e** demos is a different embodiment, action space and
+task, so a LIBERO score would measure a sim-to-sim gap, not this cell. The
+`examples/quickstart-pi05/` LIBERO eval is for a **LIBERO-compatible** π0.5
+checkpoint (e.g. the shipped `pi05_libero`). For the drug-sort model, the honest
+metric is the standalone closed-loop grasp+place eval against physics ground
+truth (`meta/gt`).
--- a/examples/quickstart-pi05-train/README.md
+++ b/examples/quickstart-pi05-train/README.md
@@ -80,6 +80,13 @@ The runner executes, in order:
 > registered config's default `repo_id`, or step 1 writes norm stats under one
 > `repo_id` and step 2 reads another and fails (loudly, after the full dataset scan).
 
+> Norm stats are cached across runs at `~/.odyssey/pi05_assets/<config_name>/<config_name>/<repo_id>/`
+> and step 1 is skipped on a hit. The hit is keyed on `config_name` + `repo_id`
+> only — it has **no content fingerprint**, so a dataset recaptured under an
+> unchanged `repo_id` would silently reuse stale statistics. When the dataset
+> content changed, set `config: {norm_stats_cache: false}` to force a fresh
+> recompute into the per-run dir.
+
 Both run with `cwd` = the task's `output_dir`, so openpi's cwd-relative `./assets`
 and `./checkpoints` land under the odyssey run dir. The trained checkpoint is
 captured from `checkpoints/<config_name>/<exp_name>/<step>/` (highest step).
@@ -94,6 +101,7 @@ captured from `checkpoints/<config_name>/<exp_name>/<step>/` (highest step).
 | `overwrite`          | emit `--overwrite` (default `true`); clobbers the prior exp dir    |
 | `resume`             | emit `--resume` instead of `--overwrite`; continue latest ckpt    |
 | `compute_norm_stats` | run the norm-stats pre-step (default `true`)                       |
+| `norm_stats_cache`   | reuse cached norm stats across runs, keyed by `config_name`+`repo_id` (default `true`); set `false` to force a recompute when the dataset content changed under the same `repo_id` |
 | *anything else*      | forwarded as a tyro override (`a_b` → `--a-b`; nested `x: {y: 1}` → `--x.y 1`; bools → `--flag` / `--no-flag`) |
 
 ## Can I then evaluate on LIBERO?
--- a/examples/quickstart-pi05-train/README.md
+++ b/examples/quickstart-pi05-train/README.md
@@ -70,10 +70,15 @@ odyssey run mission.yaml
 
 The runner executes, in order:
 
-1. `python scripts/compute_norm_stats.py pi05_ur10e_drugsort`
+1. `python scripts/compute_norm_stats.py --config-name pi05_ur10e_drugsort`
    (skip with `config: {compute_norm_stats: false}` if `./assets` already holds them)
 2. `python scripts/train.py pi05_ur10e_drugsort --exp-name finetune-pi05-ur10e --overwrite \
       --data.repo-id ur10e_partial_cond_aug --num-train-steps 30000 --batch-size 32`
+
+> The norm-stats step takes no `data.*` overrides — it reads the dataset fixed by
+> the config's `data` factory. So the mission's `data.repo_id` **must equal** the
+> registered config's default `repo_id`, or step 1 writes norm stats under one
+> `repo_id` and step 2 reads another and fails (loudly, after the full dataset scan).
 
 Both run with `cwd` = the task's `output_dir`, so openpi's cwd-relative `./assets`
 and `./checkpoints` land under the odyssey run dir. The trained checkpoint is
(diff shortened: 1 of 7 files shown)


Danny024 (dismissed):
Reviewed the incremental range `7894913..1d534dc`.

**The fix is correct.** I verified the `--config-name` claim three independent ways against upstream openpi:

| Source | Evidence |
|---|---|
| `scripts/compute_norm_stats.py` | ends `tyro.cli(main)` with `def main(config_name: str, max_frames: int \| None = None)` — tyro exposes non-positional-only params as `--kebab-case` flags |
| openpi `README.md:154` | `uv run scripts/compute_norm_stats.py --config-name pi05_libero` |
| openpi `training/data_loader.py:179` | error text says ``Make sure to run `scripts/compute_norm_stats.py --config-name=<your-config>` `` |

The asymmetry your docstring describes is real: `train.py` goes through `tyro.extras.overridable_config_cli(...)`, which makes the config name a subcommand — positional. The docstring is accurate.

Ran the PR head locally in a fresh venv: **ruff clean, mypy strict clean (79 files), 525 passed / 4 skipped**. CI green on 3.10/3.11/3.12.

---

### One thing to fix in this PR

The commit corrected the code and the local docstring, but four other places still document the old positional form — including the copy-pasteable commands, which reproduce the exact exit-2 this fixes:

- `src/odyssey/runners/models/pi05_train.py:17` — "positional arg to **both scripts**" (contradicts the new docstring at :195)
- `src/odyssey/runners/models/pi05_train.py:28` — ``scripts/compute_norm_stats.py <config_name>``
- `examples/quickstart-pi05-train/README.md:73` — `python scripts/compute_norm_stats.py pi05_ur10e_drugsort`
- `examples/quickstart-pi05-train/README.md:87` — "`config_name` … (positional)"
- `examples/quickstart-pi05-train/setup.sh:143` — `python scripts/compute_norm_stats.py $CONFIG_NAME`

All five are one-line edits.

### Suggestion: a conformance test would catch the next one pre-GPU

Two commits on this PR are upstream-CLI-contract fixes found by running on a GPU. `tests/conformance/test_gr00t_upstream.py` already exists for exactly this — its docstring says it pins the CLI surface against upstream source "so drift shows up as a red test instead of a failed training run on a rented GPU". It's env-gated, `ast`-based, no install or GPU needed.

A `test_pi05_upstream.py` gated on `OPENPI_REPO_PATH` asserting (a) `compute_norm_stats.main`'s params aren't positional-only, (b) `train.py` routes through `overridable_config_cli`, (c) `TrainConfig` carries the fields the example overrides — would have caught this one. Happy for it to be a follow-up rather than a blocker.

### Minor: norm-stats gets no overrides, train does

`build_pi05_norm_stats_argv` passes only `--config-name`, while `build_pi05_train_argv` forwards `data.repo_id`. If a registered config's default `repo_id` ever diverges from the mission override, step 1 writes to `assets/<config>/<default>` and step 2 reads `assets/<config>/<override>`. I traced it — `_load_norm_stats` swallows `FileNotFoundError` → `None`, then `data_loader.py:176` raises, so it fails loudly, just after the full dataset scan. One README line ("the mission's `data.repo_id` must equal the registered config's") closes it.

### Contract claims I confirmed against upstream source

These can come off the "unverified" list in the PR description:

- `train.py <config_name>` positional ✓
- `--exp-name` ✓ (note `train_pytorch.py` uses `--exp_name` — the runner targets the right script)
- `overwrite` / `resume` are mutually exclusive upstream (`config.py:555` raises if both), so the runner's `if resume … elif overwrite` is correct ✓
- `assets_base_dir="./assets"` and `checkpoint_base_dir="./checkpoints"` are cwd-relative, so the `cwd=<task output_dir>` approach works as described ✓
- `checkpoint_dir = ./checkpoints/<config>/<exp_name>` matches `_resolve_output_checkpoint` ✓

### Housekeeping

- The description still says "no GPU smoke yet", but the commit message says "Found on the first GPU run" — worth updating to say where the run now stands.
- Description says 524 passed / 5 skipped; head is 525 / 4.

---

Approving. The change is right and the reasoning is well recorded — the commit message captures both the symptom (exit 2) and where it was found, and the test's exact-list assertion quietly pins the "norm-stats takes no overrides" contract as well as the flag name. Just the doc lines, please, then it's good to merge.

SoyGema:
@Danny024 thanks for the approve and the thorough upstream cross-check — pushed `055233d`.

**The one fix you asked for (all 5 doc lines):**
- `pi05_train.py:17` — replaced "positional arg to both scripts" with the real asymmetry: `train.py` takes `config_name` positionally (`overridable_config_cli` subcommand), `compute_norm_stats.py` takes it as a `--config-name` flag (plain `tyro.cli`).
- `pi05_train.py:28` — `compute_norm_stats.py --config-name <config_name>`.
- `README.md:73` — the run-order command.
- `README.md` config table row — `config_name` now notes positional-to-train / flag-to-norm-stats.
- `setup.sh:143` — the watch-list command.

Confirmed no other stragglers (`grep` of the old positional form across `src/` + `examples/` is clean).

**Also took the two non-blocking suggestions that were cheap:**
- Added the README note you proposed: norm-stats takes no `data.*` overrides, so the mission's `data.repo_id` must equal the registered config's default `repo_id`, else step 1 and step 2 disagree and it fails after the scan.
- Housekeeping: PR description updated to **525 passed / 4 skipped**, and the "known gap" reworded — a GPU run *did* surface the CLI-contract fixes (now verified vs upstream); what's still open is a full fine-tune-to-convergence run.

**Left as a follow-up (as you offered):** the `test_pi05_upstream.py` conformance test gated on `OPENPI_REPO_PATH`, mirroring `test_gr00t_upstream.py`. Happy to open it as its own PR so this one can merge.

Docs-only change; ruff + mypy strict clean, `test_pi05_train.py` 25 passed.

Danny024 (changes requested):
Re-reviewed after the two new commits. My earlier approval was auto-dismissed by the push (`dismiss_stale_reviews_on_push` is on for `develop`), so this needs a fresh one either way.

**`055233db` — docs fix: correct.** All five stale spots I flagged are fixed, and the remaining "positional" mentions are now accurate because they scope it to `train.py`. The README table row reads well.

**Verified on `fe0fe6ec`:** ruff clean · mypy strict clean (79 files) · **527 passed, 4 skipped** · 27 pi05 unit tests.

---

## One blocking issue in the cache commit

`fe0fe6ec` keys the cache by `config_name`, but **openpi keys norm stats by `repo_id`**. I checked this against the artifacts from your real GPU run rather than inferring it:

```
.odyssey/runs/6b819a87…/checkpoints/pi05_ur10e_drugsort/finetune-pi05-ur10e/29999/
    assets/ur10e_partial_cond_aug/norm_stats.json
          ^^^^^^^^^^^^^^^^^^^^^^^ the dataset, one level inside assets
```

So `any(cache.rglob("norm_stats.json"))` reports a hit whenever *any* dataset has stats under that config — not the one this run needs. Reproduction against the branch:

```python
def test_cache_hit_is_blind_to_the_dataset(tmp_path, monkeypatch):
    monkeypatch.setenv("HOME", str(tmp_path / "home"))
    CFG = "pi05_ur10e_drugsort"
    run1 = tmp_path / "run1"; run1.mkdir()
    cache, _ = _link_norm_stats_cache(run1, CFG)
    (cache / CFG / "dataset_A").mkdir(parents=True)
    (cache / CFG / "dataset_A" / "norm_stats.json").write_text("{}")

    run2 = tmp_path / "run2"; run2.mkdir()
    _, cached2 = _link_norm_stats_cache(run2, CFG)     # run2 uses dataset_B
    assert cached2 is False                            # FAILS: reports True
```

The consequence is a confusing failure rather than a clean one: the runner logs *"reusing cached norm stats"*, skips the step, and then `train.py` cannot find stats for the actual `repo_id` and dies at openpi's `data_loader.py:176` — an error that says norm stats are missing, immediately after we said they were cached.

It's reachable from the shipped example, which sets `data: {repo_id: ur10e_partial_cond_aug}` as a tyro override. Iterating datasets against one registered config is the natural workflow, and it's the one this breaks.

The fix is small — check the specific path instead of globbing:

```python
cached = (cache / config_name / repo_id / "norm_stats.json").exists()
```

`repo_id` is available from the same place `_lerobot_env_for_dataset` gets it (`task.dataset.ref` basename), or from the `data.repo_id` override.

## A second concern worth deciding on deliberately

**The cache has no invalidation key.** Same `config_name` *and* same `repo_id`, but the dataset content has changed — recaptured, more episodes merged — is a cache hit, and stale statistics are reused with no error and no log line saying anything is stale. That trains a model on wrong normalisation and looks completely healthy.

That's the case I'd actually worry about, because it's how our datasets evolve in practice: the folder name stays, the contents grow. A mtime or episode-count fingerprint in the cache key would close it. If you'd rather not carry that complexity, an explicit `config: {norm_stats_cache: false}` escape hatch plus a warning line would at least make it recoverable.

## Smaller notes, non-blocking

- The new test only exercises same-config/same-dataset reuse, so it encodes the current behaviour as correct. Varying `repo_id` between the two runs would have caught the above.
- If `output_dir/assets` already exists as a real directory, the symlink is silently not created and caching silently does nothing. Graceful, but a debug log line would save someone a puzzled hour.
- Two concurrent runs sharing a `config_name` write into the same cache directory. Probably fine today; worth knowing before this runs as a service.
- `Path.home()` with `monkeypatch.setenv("HOME", …)` works on POSIX but not Windows. Only matters if CI ever runs there.

---

Requesting changes only for the `repo_id` key — everything else is a judgement call I'm happy to leave with you. The docs fix and the underlying training path both look right, and the caching idea is a good one: recomputing deterministic stats on every run is real wasted minutes.

SoyGema:
@Danny024 fixed the blocking `repo_id` key issue — pushed `0fb791f`. Good catch, and thanks for verifying it against the real GPU artifacts rather than inferring.

**The blocking fix — cache keyed by the exact dataset:**
- `_link_norm_stats_cache` now takes `repo_id` and checks the exact `cache/<config_name>/<repo_id>/norm_stats.json` path instead of `any(rglob(...))`. A cached `dataset_A` no longer satisfies a run for `dataset_B` under the same config.
- `repo_id` is resolved by a new `_dataset_repo_id(task)` — the `data.repo_id` tyro override when present, else the local dataset folder basename (same source as `_lerobot_env_for_dataset`). Empty `repo_id` (config default, unknown to the runner) → recompute rather than risk a wrong hit.
- Added your exact reproduction as `test_link_norm_stats_cache_hit_is_keyed_by_dataset` (varies `repo_id` between two runs, asserts the second is a miss), plus an empty-repo_id-recomputes test and three `_dataset_repo_id` tests. Rewrote the existing reuse test to the real `<config_name>/<repo_id>` layout.

**Second concern (stale content, unchanged `repo_id`) — escape hatch, as you offered:**
- Added `config: {norm_stats_cache: false}` — skips the shared cache entirely and recomputes into the per-run assets dir, with a `logger.warning`. Documented in the README (table row + a note that the cache has no content fingerprint). Went with the escape hatch over a mtime/episode-count fingerprint to keep the key simple; happy to revisit the fingerprint later if it bites.

**Smaller notes:** the new dataset-varying test covers the first bullet you raised. The symlink-already-a-real-dir and concurrent-run notes I left as-is for now (both benign today) — can add a debug line in a follow-up.

Verified: ruff clean · mypy strict clean (79 files) · **531 passed, 5 skipped** · 32 pi05 unit tests. Re-requesting review.

Danny024 (approved):
Approving — `0fb791f` fixes the blocking issue and takes the non-blocking one with it.

## The cache key

```python
cached = bool(repo_id) and (cache / config_name / repo_id / "norm_stats.json").is_file()
```

Exact path instead of the recursive glob. I re-ran my reproduction against this head and it now passes on all three branches that matter:

| scenario | before | now |
|---|---|---|
| same config, same dataset | hit | **hit** — caching still does its job |
| same config, **different** dataset | hit (the bug) | **miss** |
| unknown `repo_id` | hit | **miss** — recomputes rather than guessing |

That last one is the right call and I hadn't asked for it: when neither `data.repo_id` nor a local dataset ref is available, the registered config's default isn't visible to the runner, so a hit would be a guess. Recomputing costs minutes; a wrong hit costs a training run.

`_dataset_repo_id`'s precedence — override, then folder basename, then empty — matches where openpi actually gets the value, and reuses the same source as `_lerobot_env_for_dataset` rather than inventing a second path.

## The escape hatch

`norm_stats_cache: false` addresses the concern I flagged as a judgement call rather than a blocker: same config, same `repo_id`, changed dataset *contents* would silently reuse stale statistics. Bypassing the shared cache entirely and recomputing into the per-run dir is the right shape — simpler than a content fingerprint, and it fails toward recomputation. The `logger.warning` matters too, since a silent bypass would be its own small mystery later.

Correctly added to `_CONTROL_KEYS`, so it isn't forwarded to tyro as a flag.

## Verified

ruff clean · mypy strict clean (79 files) · **531 passed**, up from 527, with seven new cache tests covering the dataset-keyed hit, the empty-`repo_id` recompute, and all three `_dataset_repo_id` resolution paths.

(The one failure on my machine is `test_gr00t_isaac_eval.py::test_build_server_command_defaults_and_omits_optional`, which fails identically on `develop` and passes in CI — ambient environment, not this PR. Flagged separately on #88.)

## Not blocking, carried forward from the first review

`compute_norm_stats.py` still accepts no field overrides, so a mission whose `data.repo_id` differs from its registered config's will compute stats under one id and train looking for another. The README line you added covers it and the failure is loud rather than silent, so this is documentation-adequate — worth knowing it's a constraint of openpi's CLI rather than something this runner can fix.

Good to merge. Nice catch adding the unknown-`repo_id` case.

Issue #89, opened 2026-08-03
eval harness mutates the physics at runtime and the cell XML doesn't say so


Danny024:
## What

`examples/ur5e-drugsort/browser_capture/eval_browser_groot.js` changes the physics at runtime, and the cell XML does not record it. Two places:

**`SETUP()`** disables collision on every arm and gripper geom except the pads:

```js
if ((ARM_BODIES.has(bn) || bn.startsWith('gr_')) && !gn.includes('pad')) {
  m.geom_contype[g] = 0; m.geom_conaffinity[g] = 0;
}
```

**`RUN_ATTEMPT()`** raises the solver's no-slip iterations:

```js
m.opt.noslip_iterations = 20;
```

## Why it matters

The physics the eval runs is not the physics in the cell file, so a cell that is correct on disk can behave differently under eval, and analysis done against the XML studies a different world than the numbers came from.

Concretely, this bit us: the UR10e cell has a `shoulder_link` collision capsule 27 mm through the world ground plane (separate issue). Under eval that contact is silently absent because arm collisions are off; in native MuJoCo it produces a 1.42e17 N contact that swamps every force reading. Two of my diagnostic probes were built on the wrong assumption before I found this.

It also means the eval cannot observe arm self-collision or arm-environment collision at all — an arm may pass through the rack it is working in and nothing objects.

## Suggested fix

Move both settings into the cell XML (a `<contact><exclude>` block and `<option noslip_iterations="20">`), or emit them into the eval summary so a reader of the results knows which physics produced them. The general principle: the scene file should be the single description of the world.

Found while diagnosing UR10e drug-sort placement accuracy.

SoyGema:
Fix applied on both branches that carry `eval_browser_groot.js` (the file is not on `develop`, so this went straight onto the feature branches).

## What we did

We took the **fallback** from the "Suggested fix" — *emit the overrides into the eval summary so a reader knows which physics produced the numbers* — because the ur5e/ur10e cell XML lives in the external `robot-playground` app, not in this repo, so the "move it into the scene XML" option is not actionable from here.

**Zero behaviour change** — same geoms disabled, same `noslip=20`. The mutation is now *traceable* instead of hidden:

- Hoisted the no-slip value into a named `NOSLIP_ITERATIONS` constant, passed via `cfg.noslipIterations` (no more magic `20` buried in `RUN_ATTEMPT`).
[...]

PR #92, opened 2026-08-04
feat(eval): browser closed-loop eval harness with in_well placement metric

Danny024:
Lands the browser closed-loop eval harness on `develop`, including @SoyGema's #89 fix.

## Why now

`eval_browser_groot.js` has never existed on `develop` — it has only ever lived on feature branches. That is the root of two problems: the metric correction below was invisible to everyone not on my branch, and #89 had to be fixed twice, once per branch, because there was no shared copy to fix.

The branch it currently lives on (`feat/gr00t-ur5e-residual-rl`) is being deleted — it accumulated 141 unreviewed commits and ~2,800 committed `node_modules` files. This PR is the part worth keeping.

## The metric is the substance

The historical placement criterion was `seated` — within **5 cm** laterally. The task needs **4 mm**. A 5 cm tolerance passes an object resting on the rack surface next to the hole, so `seated` was measuring something the task does not care about.

Re-scoring a 40-episode ground-truth set against the added `in_well` criterion (4 mm lateral, base below the rim, upright): **25/40 → 7/40**, a 45-point overstatement. `seated` is retained so older results stay comparable, but `in_well` is the number to quote, and the README says so.

Tilt is measured **at release**, not at rest. An object that topples off a rim and settles upright is not a successful placement, and measuring at rest hides the cause.

## Contents

| Commit | Author | What |
|---|---|---|
| `97ef999` | @Danny024 | The harness + a README documenting prerequisites, config, and the metric distinction |
| `00f8456` | @SoyGema | #89 — record the runtime physics overrides in the summary (cherry-picked from `df000b0`) |
| `93bf9a4` | @Danny024 | Refer to the cell XML generically rather than by deployment name |

## On #89

I verified @SoyGema's fix against my copy of the file, since she asked. Every marker of the tilt/`in_well` work survives the rebase at identical counts, `node --check` passes, and the plumbing is coherent end to end: constant → `SETUP` capture → `RUN_ATTEMPT` consumption → summary + stdout.

This is static verification only — a live run needs the GPU box, Chrome, the served scene and a checkpoint.

**#89 should stay open.** This makes the override honest, it does not remove it. The eval still cannot observe arm self-collision or arm-environment contact, so a defect in the scene's arm collision geometry is invisible here. Removing the override is a cross-repo change.

One small note for a follow-up, not worth blocking on: `RUN_ATTEMPT` keeps a hard-coded `: 20` fallback while the summary always reports `NOSLIP_ITERATIONS`. The fallback is necessary — it runs in the page context and cannot see the Node-side constant — and the only call site always passes cfg, so they agree today. But if the constant changes, the summary would report the new value while a cfg-less call ran 20.

## Not included

Deliberately scoped to the harness. The rest of that branch — the cell description, the pipeline notebook, the planning docs, the media — is being moved to a private repo instead, and the `node_modules` is being deleted with the branch. #91 adds `node_modules/` to `.gitignore` so it cannot recur.

## Verification

- `node --check` passes on the final file
- No `node_modules`, no host addresses, no checkpoint paths, no deployment names in the diff

### DIFF
--- a/examples/ur5e-drugsort/browser_capture/README.md
+++ b/examples/ur5e-drugsort/browser_capture/README.md
@@ -1,0 +1,67 @@
+# Browser closed-loop eval harness
+
+`eval_browser_groot.js` scores a pick-and-place policy in the deployment
+renderer rather than in a native physics viewer, because the two do not agree:
+policies score differently the moment the pixels come from a different
+renderer, so the eval has to run where the policy will run.
+
+It drives a headless Chrome against the served scene, replays a frozen seeded
+fold of object start poses, queries the policy over HTTP, and writes a summary
+plus per-episode telemetry.
+
+## Prerequisites
+
+This harness is the scoring half of a stack it does not contain:
+
+| Needs | Provided by |
+|---|---|
+| The scene served over HTTP on `$PORT` | the deployment renderer app (external) |
+| `plans_eval.json` — the frozen fold | `precompute_plans.py --seed <fold>` |
+| A policy endpoint | your serving stack |
+| `puppeteer-core` + a Chrome binary | `npm install`; `$CHROME` |
+
+Run `npm install` in this directory — `node_modules/` is gitignored and must
+never be committed.
+
+## Configuration
+
+All via environment: `PLANS`, `OUT`, `PORT`, `AGENTS` (`groot` |
+`groot-observer`), `N`, `N_ACTION_STEPS`, `MAX_TICKS`, `CHROME`, `ARM`,
+`RAW_DUMP=1` to dump policy frames and ground truth for observer retraining.
+
+## Placement metrics — read this before quoting a number
+
+The harness emits three placement criteria per episode. They are not
+interchangeable, and the difference is large enough to invert a conclusion.
+
+| Field | Criterion | Use |
+|---|---|---|
+| `success` | lifted, then released near the target | coarse |
+| `seated` | within 5 cm laterally | **legacy — loose** |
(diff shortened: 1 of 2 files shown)


SoyGema (approved):
Approving. This is the part of `feat/gr00t-ur5e-residual-rl` worth keeping, and it's cleanly scoped to `examples/ur5e-drugsort/browser_capture/` — no changes to the `odyssey` package, which is why CI is green without exercising the JS.

Verified the premise: `eval_browser_groot.js` does not exist on `develop` (`git ls-tree -r origin/develop | grep browser_capture` is empty), so rescuing it here is the right move. The #89 override plumbing is coherent end to end — constant → `SETUP` capture → `RUN_ATTEMPT` consumption → `physics_overrides` in the summary + `PHYSICS_OVERRIDES` on stdout.

The metric change is the substance and it's sound: `in_well` (4 mm, base below rim, upright) vs the legacy `seated` (5 cm), with the geometry derived from the cell XML rather than assumed, and `seated`/`success` kept unchanged so historical folds stay comparable. Additive only, no data regression.

Two minor, non-blocking notes:
- The `: 20` hardcoded fallback for `noslip_iterations` in `RUN_ATTEMPT` (page context can't see the Node-side `NOSLIP_ITERATIONS`). Fine today since the only call site always passes cfg, but if the constant ever changes, a cfg-less call would run 20 while the summary reports the new value. Worth a small follow-up, not a blocker.
- Agree #89 should stay open — this makes the override honest, it doesn't remove it, and the eval still can't observe arm self/environment collision.

Nice work on the README documenting the metric distinction. LGTM.

PR #94, opened 2026-08-05
release: v0.1.0a5 (develop → main)

SoyGema:
Promotes `develop` to `main` for the **v0.1.0a5** alpha release.

## What ships (since a4)

| PR | What |
|---|---|
| #86 | ur10e embodiment (GR00T NEW_EMBODIMENT) |
| #88 | `CustomEvalRunner` for `evaluation_type: custom` |
| #90 | π0.5 (openpi) training runner (`Pi05Runner`) + fine-tune example |
| #91 | ignore `node_modules/` |
| #92 | browser closed-loop eval harness + `in_well` placement metric |
| #93 | version bump to 0.1.0a5 |

## Merge mechanics (per the release recipe)

- Version bumped to `0.1.0a5` on **develop first** (#93), so `main` never takes a direct version commit.
- This branch is `main` + `git merge --no-ff -X theirs origin/develop`. **`git diff --stat HEAD origin/develop` is empty** — the tree is byte-identical to develop, the safety proof that `-X theirs` only dropped stale main-only content the release supersedes.
- The merge commit carries both `main` and `develop` as parents → heals ancestry, **no back-merge needed**.

⚠️ **Merge this as "Create a merge commit", NOT squash** — squash is the root cause of the divergence cycle this recipe exists to avoid.

## Verification (release branch, local)

- `git diff --stat HEAD origin/develop` → empty (tree == develop)
- ruff clean · mypy strict clean (80 files) · **552 passed, 5 skipped**

## After merge

- Tag `v0.1.0a5` on the merge commit.
- Backfill the missing `v0.1.0a4` tag on `198fd88` (the a4 release merge was never tagged).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,11 @@ dist/
 wheels/
 *.whl
 
+# Node deps. Example harnesses that drive a browser (puppeteer et al.) install
+# these locally; a single `npm install` is ~2,800 files, so an accidental
+# `git add -A` in an example dir swamps the repo. Lockfiles stay tracked.
+node_modules/
+
 # Virtual envs
 .venv/
 venv/
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -109,7 +109,7 @@ and that `agents` contains exactly one agent.
 
 | Form | Example | What it does today |
 |---|---|---|
-| `embodiment:` | `franka_panda`, `ur5e`, `sawyer` | Names a built-in catalog embodiment. 8 names accepted: `franka_panda`, `panda`, `sawyer`, `iiwa`, `jaco`, `kinova_gen3`, `ur5e`, `baxter` — the arms Robosuite's built-in robot models cover. Resolved at mission-creation by `LocalRobotProvider`; passed through to `robosuite.make(robots=...)` at evaluation. |
+| `embodiment:` | `franka_panda`, `ur5e`, `ur10e` | Names a built-in catalog embodiment. 9 names accepted: `franka_panda`, `panda`, `sawyer`, `iiwa`, `jaco`, `kinova_gen3`, `ur5e`, `baxter` — the arms Robosuite's built-in robot models cover, passed through to `robosuite.make(robots=...)` at evaluation — plus `ur10e`, driven by the **GR00T runner** (`NEW_EMBODIMENT` finetune), which Robosuite doesn't ship: it resolves and validates, but a Robosuite eval task against it raises rather than passing through. All resolved at mission-creation by `LocalRobotProvider`. |
 | `urdf:` | `./arms/my_arm.urdf` | Names a local URDF/xacro path. Existence-checked at mission-creation. No robot pass-through to Robosuite — falls back to the env's default robot. |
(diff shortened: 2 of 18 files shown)


Danny024 (approved):
Approving. I verified each mechanical claim rather than reading them, because `-X theirs` is the one strategy that can silently discard work — so the safety proof deserved reproducing.

## Verified

| Claim | Result |
|---|---|
| Tree byte-identical to `develop` | `git diff --stat <head> origin/develop` empty |
| Merge commit carries both parents | `parent1=ce88249` (main), `parent2=70124dd` (develop) — ancestry heals, no back-merge needed |
| `-X theirs` only dropped stale main-only content | The only main-only commits are the a4 release merges, and they introduce **zero** unique file content. Nothing was discarded. |
| Version bumped on develop first | `0.1.0a5` consistent in both `pyproject.toml` and `src/odyssey/__init__.py` |
| ruff / mypy strict / 552 tests | CI green on 3.10, 3.11, 3.12 |
| `v0.1.0a4` untagged | Confirmed |

Public-repo hygiene also clean: no `node_modules` in the release tree, no secrets, no absolute paths, no platform code.

## Four notes, none blocking the merge

**1 — The merge-commit warning is unenforced, and that is the actual risk.** You are right that squash is the root cause of the divergence cycle, but the repo still has `allow_squash_merge: true`, so nothing stops the next person picking it from the dropdown. A squash produces a commit with no parent link to `develop`, and divergence returns silently.

**Correction to my original text here:** I said I had disabled squash merging. I tried, and could not — my token has `push` but not `admin` on this repo, so the setting is unchanged and squash is still selectable. Apologies for the momentarily inaccurate note.

Someone with admin rights needs to run:

```bash
gh api repos/lovellai-dev/odyssey -X PATCH -F allow_squash_merge=false
```

Worth considering `allow_rebase_merge=false` alongside it — rebase merging also produces no merge commit, so it breaks ancestry exactly the way squash does. With both off, "Create a merge commit" becomes the only option and the recipe is enforced by the repo rather than by everyone remembering the warning in this description.

Until then the ⚠️ in your description is the only thing protecting the merge — so please do pick "Create a merge commit" by hand.

**2 — Two tags are missing, not one.** `v0.1.0a2` is also absent — `a1` and `a3` are tagged, `a2` and `a4` are not. `0.1.0a2` did ship (it appears in `44cb7ac`). Worth backfilling both while you are in the tagging step.

**3 — `examples/quickstart-pi05-train/setup.sh:69` clones openpi unpinned.** A quickstart inside a *tagged release* that tracks upstream `main` will break silently whenever Physical-Intelligence changes an interface. Pinning a commit or tag would make the release reproducible. The script is otherwise well built — `set -euo pipefail`, idempotent, and it fails loudly with useful guidance when `uv sync` breaks.

**4 — Cosmetic:** `src/odyssey/runners/evals/robosuite.py:11` names an internal repo in a comment on a public repo. Pre-existing, not introduced here.

Nice work on the release discipline — bumping the version on `develop` first so `main` never takes a direct version commit is the detail that makes this repeatable.

Issue #78, opened 2026-07-22
Calibrate the SPECIALIST completion check on LIBERO sim renders (all-cap, never completion)


SoyGema:
## Problem

In the multi-agent arms, the SPECIALIST completion check (`check_done`, Gemma-4-E2B int4) is **near-always `NO`** on LIBERO sim renders, so phases advance only by the `phase_max_steps` **cap**, never by `completion`. The semantic gating — the whole point of the closed loop, and the differentiator between the planner / delegation / orchestration arms — currently **adds no value** (100% `cap`), i.e. it runs effectively open-loop.

The eval `success_rate` is unaffected (it comes from LIBERO's own ground-truth success predicate, not the check; tasks still succeed 10/10 on `libero_object`). The problem is the *coordination* quality, and the planner-vs-delegation-vs-orchestration comparison measuring a degenerate regime.

## What we tried — consolidated (full chronology in the comments)

| # | Change | Result |
|---|---|---|
| 1 | Concrete achieved-criteria prompt (grasp = holding/lifted; place = at target) | all `NO` → `cap` |
| 2 | Direct-question phrasing ("Has the robot completed \<action\>?") | all `NO` → `cap` |
| 3 | Wrist camera (`--check_image_key eye-in-hand`) | **worse** — occluded blob at grasp; reverted to agentview |
| 4 | Raw VLM reply surfaced on the trace | clean `'NO'` → **not a parser bug** |
| 5 | Zoom + upscale the check frame (`check_crop 0.6`, `check_upscale 384`) | grasp now clearly visible to a human/strong VLM, **judge still `NO`** → rules out scale/framing |
| 6 | **Standalone isolation PoC** — exact `check_done` prompt (verified char-for-char) + `_parse_yes_no`; agentview; eyeball GT | see table below |

### Standalone isolation PoC (measured)

| judge | quant | acc | false-trans | missed (recall) |
|---|---|---|---|---|
| gemma-4-E2B-it | int4 | 58% | 0 | **5/5** |
| gemma-4-E4B-it | int4 | 58% | 0 | **5/5** |
| gemma-4-E4B-it · multi-frame (W=3, S=3) | int4 | 58% | 0 | **5/5** |
| gemma-4-E4B-it | bf16 | — | — | untested (OOM on T4 16 GB → needs L4/A100) |

[...]

PR #95, opened 2026-08-10
feat(cosmos3): eval-first integration of the NVIDIA Cosmos 3 WAM family

SoyGema:
## What

Eval-first integration of **NVIDIA Cosmos 3**, the World Action Model (WAM) family (Edge 4B / Nano 16B / Super 64B). A Cosmos 3 policy answers one query with a predicted future observation **video AND an action chunk**; from Odyssey's side it is a chunk-emitting pilot behind an external server — the posture π0.5 established. Family-wide by construction: no per-checkpoint code; the checkpoint lives server-side, variant knobs travel as mission config, and the chunk size is auto-adopted from the server's `GET /info`.

The SFT **training runner is deliberately NOT in this PR** (follow-up): NVIDIA publishes only DROID policy checkpoints; the paper's LIBERO-10 numbers come from their own SFT on `nvidia/LIBERO_LeRobot_v3` (LeRobot v3 — `DatasetFormat.LEROBOT` already covers it, zero spec changes).

## How

- **`pilot: cosmos3` on LIBERO evals** — `LiberoRunner` dispatch + `build_cosmos3_libero_argv` → `cosmos3_libero_eval.py` subprocess recipe → `make_cosmos3_pilot()` (stdlib `urllib` client, no new deps) behind the shared `ChunkPilotAdapter`. Server pre-started: `python -m cosmos_framework.scripts.action_policy_server_libero --checkpoint-path <local-dir> --port 8000`. Wire: `POST /predict {image, prompt, domain_name, image_size}` → `{action: [[T×D]], video: [...]}` (de-normalization is **server-side**); `n_action_steps` omitted → auto-adopt `action_chunk_size` via `GET /info` (unreachable `/info` now logs a warning naming the assumed family default instead of failing silently). *The LIBERO **example** was trimmed to keep one canonical example; the LIBERO **code path** is fully in this PR and unit-covered.*
- **`cosmos3_transforms.py`** — concat_view (third-person | wrist) base64-PNG packer; 10-D `[dpos(3), rot6d(6), gripper(1)]` → 7-DoF decoder reusing `gr00t_transforms` kinematics; ≥7-D env-native rows pass through. `translation_only` forces the gripper open in **LIBERO's convention (`-1` = open)** — fixed in review; the `+1` was Isaac's convention carried over. `domain_name` must be a registered cosmos-framework embodiment (`droid_lerobot`, `libero`, `robomind-ur`, …) — unknown values return an empty chunk.
- **First-class `evaluation_type: robolab`** — new `EvaluationType.ROBOLAB` + `RobolabRunner` (sibling of LIBERO/Robosuite/Isaac Lab). RoboLab is NVlabs' **Isaac Lab-based sim twin of the DROID rig** — the benchmark NVIDIA scores Policy-DROID checkpoints on. Contract pinned from RoboLab source: launches `policies/<backend>/run.py` headless (`num_episodes` → `--num-runs × num_envs`, rounded **up** — 10 eps / 4 envs runs 12; documented, and the summary reports the real count), passthrough keys dashed, `--output-folder-name odyssey_<task-id>`; scores `episode_results.jsonl` into the shared summary with per-task rates + failure reasons, copies the raw jsonl as a task artifact. `config.entry_script` selects the backend (family-wide); strict-mypy-clean, no exemption.
- **`OpenAICompatCompletionJudge`** (`runners/agents/openai_judge.py`) — a `CompletionDetector` over any OpenAI-compatible `/chat/completions` VLM (motivating case: a served Cosmos 3 **Reasoner**); the stronger-judge direction for #78. Strict-mypy-clean, injectable transport, conservative NO on failures.
- **One canonical example**: `examples/cosmos3-robolab/` (idempotent `setup.sh`, hardware-validated). Two sibling examples were exercised on the same bring-up and then trimmed to keep the branch focused — recoverable from git history and referenced in the robolab README: `quickstart-cosmos3` (LIBERO `pilot: cosmos3`; `88e40c9`/`b2babcb`) and `cosmos3-droid-openloop` (open-loop GT decode check incl. the `robomind-ur` path; `4b83506`). ⚠ After a **squash merge** those SHAs aren't reachable from `develop` — the README now says to `git fetch origin pull/95/head` first (or merge this PR with a merge commit). The Reasoner SPECIALIST probe lives on the `experiment-specialist-grasp-verification` branch. mypy: `cosmos3_*` glue joins the SDK-mirror `ignore_errors` list.

## Validated on hardware (H100, 2026-08-10)

- **LIBERO smoke (OOD)**: end-to-end green — `/predict` wire, `/info` chunk adoption (16), 2 full episodes + per-episode MP4s, scored summary. 0% success expected (DROID checkpoint on LIBERO).
- **Open-loop GT, in-distribution (DROID episode)**: **MAE 0.385 vs 0.946 zero-baseline** — serving + decode produce real signal. JSON-vs-plain prompts moved MAE <1%.
- **UR embodiment (for the UR10e mission series)**: `robomind-ur` accepted requests built from the team's own UR drugsort episodes and answered the correct 7-D (6 joints + gripper) rows; ≈ zero-baseline scores confirm SFT is the gate (expected, DROID-SFT checkpoint). The UR dataset was converted **LeRobot v2.1 → v3** (official converter; 160 eps / 143,449 frames, zero loss) and verified against cosmos-framework's v3-only reader.
- **RoboLab bring-up**: the docker route **dead-ended on the H100** (the nvidia toolkit never injects the driver's *graphics* libs, so Isaac's Vulkan can't start) — documented as-is; the **native uv route is now the recommended one** (`uv sync --extra isaac51 --python 3.11` + `openpi-client` + EULA env vars) and booted Isaac clean. git-lfs is mandatory (assets are LFS objects; without them task discovery reports a bogus "Task not found"); `benchmark_name` must be the task **class name**.
- **Ops findings documented in the branch**: `--checkpoint-path` requires a LOCAL dir; policy checkpoints are public but **`nvidia/Cosmos-Guardrail1` is gated** (guardrails default ON → startup crash; HF-access or disable-guardrails remedies documented); port/VRAM collisions on shared boxes documented (audit `nvidia-smi` co-tenants).

## Review follow-ups (`4f0fa81`)

Addressed from the first review: LIBERO gripper polarity in `translation_only` flipped to `-1` (open) + test assertion; `Cosmos3HttpClient.info()` wraps `URLError` like `infer()` and the `/info` fallback logs a warning with the assumed chunk size; `build_robolab_argv` ceil rounding documented; README squash-recovery note added.

## Testing

- 58 unit tests across `-k "cosmos3 or openai_judge or robolab"`: transforms round-trips, /predict contract, rot6d decode vs shared kinematics, fake-client pilot wiring incl. /info adoption, LiberoRunner dispatch/argv/subprocess bridge, robolab enum/registry/argv/results/runner, judge parsing/failure/auth/gate integration, stdlib-import cleanliness.
- Full suite green (610 passed); `mypy` strict clean; `ruff` clean.
- **In flight**: RoboLab smoke through the new runner (`ToolOrganizationTask`, Edge-Policy-DROID served over WebSocket) on the native route — in-distribution, real success rate + sim videos expected.


SoyGema:
## ✅ H100 smoke (2026-08-10) — eval path validated end-to-end

Ran the full `pilot: cosmos3` LIBERO path on the H100 box with **Cosmos3-Edge-Policy-DROID** (OOD on LIBERO by design — wiring validation, not success):

- 42/42 unit tests green on the box (py3.12) · `setup.sh` both halves green (client LIBERO venv + cosmos-framework env)
- Server live: `GET /info` → `action_chunk_size: 16` (pilot auto-adopted it), `POST /predict` wire confirmed
- Mission `COMPLETED`: 2 episodes × 120 steps, `ODYSSEY_EPISODE`/`ODYSSEY_RESULT` parsed, scored summary + per-episode MP4s

**Hardware findings folded into the branch** (commits `a87ca01`, `359edd7`):
1. `--checkpoint-path` requires a **local dir** (HF id crashes) → setup.sh now stages via `hf download`
2. Policy checkpoints are **public** (dropped wrong gated/HF_TOKEN notes)
3. `nvidia/Cosmos-Guardrail1` (guardrails default ON) **is gated** → server dies at startup without approval; documented HF-access or disable-guardrails remedies
4. Port collisions on shared boxes documented (`ss -tlnp` + `config.port`)

**Still pending** (needs the LIBERO SFT training runner, next PR): rot6d axes + gripper polarity producing real successes with an in-distribution checkpoint.

SoyGema:
## ✅ RoboLab smoke COMPLETE — first-class `evaluation_type: robolab` validated end-to-end on hardware (H100, 2026-08-11)

`ToolOrganizationTask` ("Put the red hammer and black hammer in the left bin") through the full stack: `odyssey run` → `RobolabRunner` → RoboLab/Isaac Sim (native `uv sync --extra isaac51` env) → cosmos-framework WS policy server (Edge-Policy-DROID, 69.7 ms avg inference) → `episode_results.jsonl` scored into the mission summary + **sensor & viewport MP4s** per episode.

**Result**: 0/2 success — an honest *model* result, not a pipeline one: the policy manipulates (grasps, lifts, drops) but repeatedly grabs the **cordless drill instead of the hammers** (`WRONG_OBJECT_GRABBED ×29`) — a semantic failure of the 4B Edge checkpoint on a "moderate" semantics task. The runner surfaces exactly this: per-task rates, aggregated failure reasons, kinematic metrics.

**Bring-up findings documented in the branch** (`c4f3b83`, `2b73beb`): the docker route dead-ends on hosts whose nvidia toolkit doesn't inject driver graphics libs (Vulkan can't start); the **native route is now the recommended one** — `uv sync --extra isaac51 --python 3.11` + `openpi-client` + EULA env vars + **git-lfs mandatory** (scene assets are LFS pointers; without them USD fails and task discovery reports a misleading "Task not found") + task **class name** in `benchmark_name`. Full symptom→cause→fix table in the example README.

SoyGema:
https://github.com/user-attachments/assets/07a73a24-2994-427f-a0ca-f81e9d66177a

Danny024 (changes requested):
Really nice piece of work — the eval-first framing, the family-wide-by-construction posture, and validating with a real number (MAE 0.385 vs 0.946 zero-baseline) rather than "it works" all make this easy to reason about. One correctness defect I'd like fixed before approving, then a robustness gap and two housekeeping items.

## Must fix — inverted gripper polarity for LIBERO

`src/odyssey/runners/evals/cosmos3_transforms.py:168`

```python
if translation_only:
    action[3:6] = 0.0
    action[6] = 1.0  # gripper forced open
```

In LIBERO **`-1` is open**. Three places on `develop` agree:

- `runners/evals/libero.py::_libero_action` — *"positive = close upstream → LIBERO wants the opposite"*, and it inverts the sign
- `runners/evals/gr00t_libero_eval.py:245` — `[0,0,0,0,0,0,-1.0]  # no-op, gripper open`
- `runners/evals/pi05_libero_eval.py:209` — same, `-1.0  # no-op, gripper open`

The `+1.0` looks carried over from `gr00t_isaac_eval.py:452` (`a[6] = 1.0  # fixed-open gripper`), where it's correct — Isaac's IK-rel visuomotor convention is `+1` open, the opposite of LIBERO's. It's a cross-convention copy landing in a function named `cosmos3_action_to_libero`.

`tests/unit/test_cosmos3_pilot.py:172` asserts `action[6] == pytest.approx(1.0)  # gripper forced open`, so the test encodes the same inversion — green CI can't catch this. Both need flipping.

Scope is the `translation_only` de-risk path, not the headline eval numbers. But that knob exists to separate translation failures from grasp failures, and with the gripper clamped shut it can never grasp — so it misleads exactly when someone reaches for it. Your own docstring warning (*"⚠ Verify gripper polarity and rot6d axis order on the first GPU rollout"*) is what sent me looking, so the instinct was right.

## Worth fixing — the `/info` chunk-size fallback is silent

`src/odyssey/runners/models/cosmos3.py:113-120`

```python
except Exception:  # /info unreachable -> family default, fail late in act()
    n_action_steps = COSMOS3_DEFAULT_CHUNK_SIZE
```

The module has no logger at all. Auto-adopting `action_chunk_size` is one of the nicer parts of the design, but when it *doesn't* work the operator gets no signal — and if the server's real chunk is 8 while the default guesses 16, it surfaces as an index error deep inside `act()`, which the comment already anticipates. One `logger.warning` naming the fallback and the assumed size turns a puzzling runtime failure into an obvious one.

Minor and related: `Cosmos3HttpClient.info()` doesn't wrap `URLError` the way `infer()` does, so it raises a raw error to a caller that only survives because of the broad `except`.

## Housekeeping

**The trimmed-example SHAs won't survive a squash merge.** `examples/cosmos3-robolab/README.md` points at commits `88e40c9`/`b2babcb`/`4b83506` to recover `quickstart-cosmos3` and `cosmos3-droid-openloop` "from git history". This repo has `allow_squash_merge: true`, so after a squash those aren't reachable from `develop` and `git show 88e40c9` fails on a fresh clone. Cheapest fix is a note to `git fetch origin pull/95/head` first; alternatively merge this one with a merge commit.

**`Co-Authored-By` lines** are on all 15 commits — repo convention is to leave them off. DCO sign-off is present on every commit, so that side is clean.

**Nit:** `build_robolab_argv` computes `--num-runs` as `ceil(num_episodes / num_envs)`, so `num_episodes: 10` with `num_envs: 4` actually runs 12 episodes. The summary honestly reports `len(episodes)` so nothing is misreported — worth a line in the docstring so a mission author isn't surprised.

## What I liked

- The mypy split is the right call: the exemption covers only the untyped-SDK mirror glue and follows the existing precedent, while the two genuinely new subsystems (`robolab.py`, `openai_judge.py`) are strict-clean with no exemption.
- `build_robolab_argv` guards the divisor and uses `ceil` — correct, and the bool-before-int ordering in the passthrough loop avoids the classic `isinstance(True, int)` trap.
- The judge takes `api_key_env` instead of inline secrets, and explicit keys override `extra_body` so a routing knob can't clobber `model`.
- stdlib-only client with an injectable transport means the whole path unit-tests on a CPU box — that's why the 56 tests are worth something.
- Deliberately leaving the SFT training runner out keeps this reviewable. Good call.

For transparency: I also chased a suspected `TrainingProcessSpec(timeout_seconds=...)` mismatch and it was a false alarm — my local checkout was behind `develop`, which has the field.

Happy to re-review as soon as the gripper sign is flipped — it's a one-line change plus the assertion.

SoyGema:
Thanks for the sharp review — the gripper catch especially. One clarification up front: `726b78f` trimmed the LIBERO **example** (`quickstart-cosmos3`) to keep one canonical example, but the LIBERO **code path** (`pilot: cosmos3` dispatch, `cosmos3_libero_eval.py`, `cosmos3_transforms.py`) is fully in the PR — so your must-fix absolutely applied. All points addressed in `4f0fa81`:

**Must fix — gripper polarity**: you're right, and your diagnosis of the origin is exactly what happened — the `+1.0` was carried over from `gr00t_isaac_eval.py`'s Isaac IK-rel convention into a LIBERO-named function. `translation_only` now forces `action[6] = -1.0` (LIBERO: −1 = open), with an inline comment naming the convention, and the test asserts `-1.0`. The docstring's "verify polarity on first GPU rollout" warning stays for the pass-through path, which is still unverified on a grasping task.

**`/info` fallback**: the module now has a logger; the fallback logs a warning naming the failure and the assumed family-default chunk size, and points at the `n_action_steps` override — so the "index error deep inside `act()`" scenario now has an obvious breadcrumb. Also fixed the related minor: `info()` now wraps `URLError` into the same actionable `RuntimeError` as `infer()`.

**Squash-unsafe SHAs**: took the cheap fix — the robolab README now notes that after a squash merge the commits need `git fetch origin pull/95/head` first. (If this lands as a merge commit instead, the note is harmless.)

**`Co-Authored-By`**: dropped from the new commit onward; left the existing 15 as-is rather than rewriting the reviewed history.

**`--num-runs` ceil nit**: documented in the `build_robolab_argv` docstring with the 10-eps/4-envs → 12 example, plus a note that the summary reports the real count.

Full suite is green (610 passed), mypy strict and ruff clean. PR description updated to match (LIBERO example-vs-code clarification, native RoboLab route, review follow-ups). Ready for re-review.

Danny024 (approved):
Re-reviewed `4f0fa81b`. All four points addressed, and the blocking one is correct. Approving.

**Gripper polarity — fixed, and I re-verified the convention rather than just the diff.**

```
-        action[6] = 1.0   # gripper forced open
+        action[6] = -1.0  # gripper forced open (LIBERO: -1 = open, +1 = close)
```

`-1` is right. Both sibling recipes on `develop` agree:

```
gr00t_libero_eval.py:245   [0,0,0,0,0,0,-1.0]  # no-op, gripper open
pi05_libero_eval.py:209    [0,0,0,0,0,0,-1.0]  # no-op, gripper open
```

and `libero.py::_libero_action` inverts for the same reason ("positive = close upstream → LIBERO wants the opposite"). The test flipped to `-1.0` too, so the assertion no longer cements the wrong convention. Spelling the convention out inline is a good call — that's exactly how the Isaac value got carried over in the first place.

**`/info` fallback — fixed, and slightly better than I suggested.** I only asked for a warning; you also wrapped `info()`'s `URLError` in the same actionable `RuntimeError` as `infer()`, which removes the asymmetry where a caller only survived because of the broad `except`. The warning naming the assumed chunk size is the part that will save someone an afternoon.

**README SHAs** — `git fetch origin pull/95/head` is the cheap correct fix. **`num_runs` rounding** — documented, which is all it needed.

CI green on 3.10–3.12.

Only outstanding item is the `Co-Authored-By` lines on all 16 commits, and I would not rebase for that — the cost outweighs the tidiness on a branch this size. Worth watching on the next one.

Nice work on the bring-up. The `robomind-ur` result — accepting our own UR drug-sort episodes and returning correct 7-D rows — turned out to be directly load-bearing for the pilot comparison on our side, and the "SFT is the gate" conclusion saved us from running a cross-pilot table that would have been all zeros.

